PR #20, opened 2025-11-25
feat(go): support snowflake:// scheme and add tests

Mandukhai-Alimaa:
## What's Changed

accept snowflake:// scheme uri and add tests.

### DIFF
--- a/go/driver_test.go
+++ b/go/driver_test.go
@@ -2825,10 +2825,8 @@ func TestSnowflakeURIScheme(t *testing.T) {
 
 				if tc.expectedErrCode != 0 {
 					var sfError *gosnowflake.SnowflakeError
-					assert.ErrorAs(t, err, &sfError, "Expected a SnowflakeError type for code %d", tc.expectedErrCode)
-					if sfError != nil {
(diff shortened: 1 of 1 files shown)


go/driver_test.go
```diff
@@ -2824,9 +2824,9 @@ func TestSnowflakeURIScheme(t *testing.T) {
 				require.Error(t, err, "Expected ParseDSN to fail for URI %q -> %q", tc.snowflakeURI, uri)
 
 				if tc.expectedErrCode != 0 {
-					sfError, ok := err.(*gosnowflake.SnowflakeError)
-					assert.True(t, ok, "Expected a SnowflakeError type for code %d", tc.expectedErrCode)
-					if ok {
+					var sfError *gosnowflake.SnowflakeError
+					assert.ErrorAs(t, err, &sfError, "Expected a SnowflakeError type for code %d", tc.expectedErrCode)
+					if sfError != nil {
```
lidavidm: You can just omit this check; if sfError is nil, then the test has already failed. You can use `require.ErrorAs` instead if you want the test to end at that point.

lidavidm (approved):

PR #22, opened 2025-11-27
chore: add CODEOWNERS

lidavidm:
## What's Changed

Add CODEOWNERS


amoeba (approved):
Looks good. This doesn't catch non-dotfiles at the root, I imagine that's that intent?

PR #37, opened 2025-12-09
feat(csharp): Add csharp interop

davidhcoe:
## What's Changed

Ports over the current state of https://github.com/apache/arrow-adbc/tree/main/csharp/src/Drivers/Interop/Snowflake, changed the namespace, updated headers and build scripts.

### DIFF
--- a/csharp/test/Interop/Resources/SnowflakeData.sql
+++ b/csharp/test/Interop/Resources/SnowflakeData.sql
@@ -1,6 +1,9 @@
 
  -- Copyright (c) 2025 ADBC Drivers Contributors
  --
+ -- This file has been modified from its original version, which is
+ -- under the Apache License:
+ --
  -- Licensed to the Apache Software Foundation (ASF) under one
  -- or more contributor license agreements.  See the NOTICE file
  -- distributed with this work for additional information
--- a/csharp/test/Interop/readme.md
+++ b/csharp/test/Interop/readme.md
@@ -2,6 +2,9 @@
 
 Copyright (c) 2025 ADBC Drivers Contributors
 
+This file has been modified from its original version, which is
+under the Apache License:
+
 Licensed to the Apache Software Foundation (ASF) under one
 or more contributor license agreements.  See the NOTICE file
 distributed with this work for additional information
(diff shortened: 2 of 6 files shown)


davidhcoe:
> For files ported from the Apache repo, we need to keep the original license header and add the attribution text (same as for other files)

I kept going through the pre-commit check until it didn't fail anymore. I would get details like:
```
Apache RAT Check.........................................................Failed
- hook id: rat
- exit code: 6

Checking licenses for C:\dev\davidhcoe\adbc-drivers-snowflake
Using Apache RAT: C:\Users\davidcoe\AppData\Local\ADBC Driver Foundry\adbc-drivers-dev\Cache\apache-rat-0.16.1.jar
Files missing ADBC Drivers Contributors copyright header:
- csharp/test/Interop/Resources/snowflakeconfig.json
Files that should not have 'This file has been modified' header:
- csharp/src/Interop/Build-SnowflakeDriver.ps1
- csharp/src/Interop/copySnowflakeDriver.sh
- csharp/src/Interop/readme.md
- csharp/test/Interop/Resources/SnowflakeConstraints.sql
- csharp/test/Interop/Resources/SnowflakeData.sql
```
and that all clears now. Is there something else I need to do?

lidavidm:
(I should update the check to make that clear)

csharp/test/Interop/readme.md
```diff
@@ -0,0 +1,54 @@
+<!--
+
+Copyright (c) 2025 ADBC Drivers Contributors
+
+Licensed to the Apache Software Foundation (ASF) under one
```
lidavidm: Ditto here

csharp/test/Interop/Resources/SnowflakeData.sql
```diff
@@ -0,0 +1,81 @@
+
+ -- Copyright (c) 2025 ADBC Drivers Contributors
+ --
+ -- Licensed to the Apache Software Foundation (ASF) under one
```
lidavidm: Ditto here

lidavidm (approved):

PR #50, opened 2025-12-30
refactor(go): use statement_rows_affected_ddl flag and remove test_rows_affected override

Mandukhai-Alimaa:
## What's Changed

Use statement_rows_affected_ddl flag and remove unneccessary override

### DIFF
--- a/go/validation/tests/test_statement.py
+++ b/go/validation/tests/test_statement.py
@@ -20,3 +20,7 @@ from . import snowflake
 
 def pytest_generate_tests(metafunc) -> None:
     return statement_tests.generate_tests(snowflake.QUIRKS, metafunc)
+
+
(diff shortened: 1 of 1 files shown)


go/pixi.toml
lidavidm: FWIW, I think it's OK not to pin it here since the lockfile pins it instead

go/validation/tests/test_statement.py
```diff
@@ -20,61 +20,3 @@
 
 def pytest_generate_tests(metafunc) -> None:
     return statement_tests.generate_tests(snowflake.QUIRKS, metafunc)
-
-
-class TestStatement(statement_tests.TestStatement):
```
lidavidm: The class has to be kept or there will be no tests discovered

PR #57, opened 2026-01-07
feat(go): remove override of split_statement use updated function

Mandukhai-Alimaa:


lidavidm:
There we go. @Mandukhai-Alimaa, can you rebase this?

PR #95, opened 2026-03-16
feat(go): support target catalog/schema in ingest

barino86:
## What's Changed

Adds support for the standard ADBC `adbc.ingest.target_catalog` and
`adbc.ingest.target_db_schema` options in the Go Snowflake driver. When set,
the driver builds a fully-qualified table identifier (e.g.
`"catalog"."schema"."table"`) for all ingestion operations including
`CREATE TABLE`, `DROP TABLE IF EXISTS`, `COPY INTO`, and row count queries.

Previously, ingestion only used the unqualified table name, requiring users to
connect directly to the target database and schema. With this change, users can
ingest into any `database.schema.table` from any connection context.

Closes #94.

### DIFF
--- a/go/bulk_ingestion.go
+++ b/go/bulk_ingestion.go
@@ -149,7 +149,7 @@ func (st *statement) ingestRecord(ctx context.Context) (nrows int64, err error) 
 
 	var (
 		initialRows int64
-		target      = quoteTblName(st.targetTable)
+		target      = st.qualifiedTableName()
 	)
 
 	// Check final row count of target table to get definitive rows affected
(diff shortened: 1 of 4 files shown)


lidavidm (commented):
LGTM; we need to enable the integration tests for this feature at https://github.com/adbc-drivers/snowflake/blob/905093301ec0291ff0473fd12722af3bf7e20d48/go/validation/tests/snowflake.py#L38-L39, but first we need to set up a secondary catalog and schema in CI for this to use

zeroshade (approved):
LGTM.

@lidavidm do you think that's a blocker on this or something we can do as a follow-up?

lidavidm:
Sorry for the long delay here - I am going to release the driver with this fix this week!

PR #99, opened 2026-03-18
feat: GeoArrow support for bulk ingestion (GEOGRAPHY/GEOMETRY)

jatorre:
## Summary

Adds geospatial column support to the Snowflake ADBC driver's bulk ingestion path. When Arrow columns carry `geoarrow.wkb` or `geoarrow.wkt` extension metadata, the driver automatically creates GEOGRAPHY or GEOMETRY columns in Snowflake and converts the data.

- Detects geoarrow columns from `ARROW:extension:name` field metadata (handles C Data Interface where Go-level extension types are stripped)
- New statement option `adbc.snowflake.statement.ingest_geo_type`: `"geography"` (default, WGS84/4326) or `"geometry"` (any SRID)
- Extracts SRID from geoarrow CRS metadata (PROJJSON or `"EPSG:NNNN"` format) for GEOMETRY columns
- Unit tests for type mapping and SRID extraction

### How it works

1. Detect geoarrow columns from Arrow extension metadata
2. Create table with native GEOGRAPHY/GEOMETRY columns
3. **COPY INTO with transform**: `TO_GEOGRAPHY($1:"geom"::BINARY, true)` converts WKB to GEOGRAPHY inline during COPY — no post-processing needed
4. For GEOMETRY columns, SRID is applied via `ST_SETSRID` if present in geoarrow metadata

### Why COPY transform?

Snowflake's COPY INTO from Parquet **cannot load WKB directly into GEOGRAPHY/GEOMETRY columns** — only CSV and JSON/AVRO support direct geospatial loading from stages ([docs](https://docs.snowflake.com/en/sql-reference/data-types-geospatial#loading-geospatial-data-from-stages)).

The initial approach used a post-COPY CTAS pattern (rename → CREATE TABLE AS SELECT with conversion → drop staging). This PR replaces that with a **COPY transform** that applies `TO_GEOGRAPHY`/`TO_GEOMETRY` in the SELECT clause of the COPY subquery, eliminating 3 SQL round-trips and a full table rewrite.

The original CTAS path is preserved as fallback for schemas without geo columns.

### Benchmark results

**COPY transform vs CTAS approach** (50K random points, median of 10 runs):

```
                       Median      P25      P75      Min      Max   Rows/sec   N
─────────────────────────────────────────────────────────────────────────────────
CTAS (old approach)     8.13s    8.11s    8.26s    7.99s    8.85s      6,150   7
COPY transform (this)   6.11s    5.81s    6.15s    5.34s    7.44s      8,183  10

Speedup (median): 1.33x  (6,150 → 8,183 rows/sec)
```

The COPY transform also had **zero transient failures** vs 3/10 for the CTAS path (fewer SQL round-trips = fewer timeout opportunities).

**End-to-end with real-world data** (Czech Republic OSM Geofabrik):

| Dataset | Rows | Throughput | Geometry type |
|---------|------|-----------|---------------|
| POIs | 465,280 | **38,119 rows/sec** | Point |
| Roads | 1,885,651 | **56,804 rows/sec** | LineString |
| Buildings | 5,014,886 | **68,611 rows/sec** | Polygon |

**At scale** (500K rows, single runs):

| Dataset | CTAS (old) | COPY transform | Speedup |
|---------|-----------|----------------|---------|
| 500K points | 29,499 rows/sec | 39,339 rows/sec | **1.33x** |
| 500K polygons | 25,189 rows/sec | 28,074 rows/sec | **1.11x** |

### Export (not in this PR)

Export/read-path geoarrow support is in a separate PR (#100). Detecting GEOGRAPHY/GEOMETRY columns on the read path is non-trivial because:
- With `GEOGRAPHY_OUTPUT_FORMAT=EWKB`, `srcMeta.Type` becomes `"binary"` (type info lost)
- With default GeoJSON format, `srcMeta.Type` is `"object"` (same as VARIANT/OBJECT)

### Context

This is part of a broader effort to add GeoArrow support across ADBC drivers. Previously opened as [apache/arrow-adbc#4114](https://github.com/apache/arrow-adbc/pull/4114), moved here per maintainer request.

## Test plan

- [x] Unit tests for `toSnowflakeType` with geoarrow extension types
- [x] Unit tests for `extractSRIDFromMeta` (PROJJSON, simple EPSG string, null, empty, invalid)
- [x] Existing `TestIngestBatchedParquetWithFileLimit` still passes
- [x] End-to-end tested against real Snowflake with points, lines, and polygons
- [x] Verified GEOGRAPHY column type created in Snowflake via INFORMATION_SCHEMA
- [x] Benchmarked COPY transform vs CTAS approach (10 iterations, median comparison)

### DIFF
--- a/go/bulk_ingestion.go
+++ b/go/bulk_ingestion.go
@@ -122,6 +122,13 @@ type ingestOptions struct {
 	//
 	// Default is true.
 	vectorizedScanner bool
+	// Snowflake type to use for geoarrow columns (geoarrow.wkb, geoarrow.wkt).
+	//
+	// Valid values are "geography" (default) and "geometry".
+	// GEOGRAPHY is always WGS84 (SRID 4326). GEOMETRY supports any SRID;
+	// the SRID is extracted from geoarrow extension metadata and applied
+	// via ST_SETSRID after COPY INTO.
+	geoType string
 }
 
 func DefaultIngestOptions() ingestOptions {
@@ -133,6 +140,7 @@ func DefaultIngestOptions() ingestOptions {
 		compressionCodec:  defaultCompressionCodec,
 		compressionLevel:  defaultCompressionLevel,
 		vectorizedScanner: defaultVectorizedScanner,
+		geoType:           "geography",
 	}
 }
 
@@ -141,7 +149,7 @@ func DefaultIngestOptions() ingestOptions {
 //
 // The Record must already be bound by calling stmt.Bind(), and will be released
 // and reset upon completion.
-func (st *statement) ingestRecord(ctx context.Context) (nrows int64, err error) {
+func (st *statement) ingestRecord(ctx context.Context, copyQ string) (nrows int64, err error) {
 	defer func() {
 		// Record already released by writeParquet()
 		st.bound = nil
@@ -209,7 +217,7 @@ func (st *statement) ingestRecord(ctx context.Context) (nrows int64, err error) 
 	}
 
 	// Load the uploaded file into the target table
-	_, err = st.cnxn.cn.ExecContext(ctx, copyQuery, []driver.NamedValue{{Value: target}})
+	_, err = st.cnxn.cn.ExecContext(ctx, copyQ, []driver.NamedValue{{Value: target}})
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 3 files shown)


jatorre:
Superseded by #117, which combines the GeoArrow ingest (#99) and export (#100) work with EWKB-based CRS readback into a single PR for review. Closing in favor of that. Discussion threads from this PR are still searchable; further review requests will continue on #117.

PR #105, opened 2026-03-31
perf(go): optimize GetObjects by reading SHOW results directly

zeroshade:
## Summary

- Reduce Snowflake round-trips for `GetObjects` at `ObjectDepthTables` and `ObjectDepthDBSchemas` when catalog is specific (non-wildcard)
- New direct path reads `SHOW TERSE` rows via `driver.Rows` and builds the `GetObjectsInfo` tree in Go, instead of collecting queryIDs and running complex SQL templates that re-read results through `RESULT_SCAN` joins
- Wildcard-catalog queries are unaffected and fall through to the existing SQL template path

## Motivation

Benchmark comparison ([report](https://github.com/columnar-tech/adbc-odbc-snowflake/blob/main/BENCHMARK_REPORT.md)) between the ADBC ODBC bridge and the official Snowflake ODBC driver showed `SQLTables` was **2.06x slower**. The root cause: `GetObjects(ObjectDepthTables)` executed 4 Snowflake queries (3 parallel `SHOW TERSE` + 1 complex SQL joining them via `RESULT_SCAN`) where the official driver uses essentially 1 query.

## Query count reductions

| Scenario | Before | After |
|---|---|---|
| Specific catalog + specific schema (`ObjectDepthTables`) | 4 queries | **1 query** |
| Specific catalog + wildcard schema (`ObjectDepthTables`) | 4 queries | **2 queries** |
| Specific catalog (`ObjectDepthDBSchemas`) | 3 queries | **1 query** |
| Wildcard catalog (any depth) | Unchanged | Unchanged |

## Changes

- **`go/get_objects_direct.go`** (new): Optimized `GetObjects` implementations that execute `SHOW TERSE` commands and read rows directly, building the result tree in Go via `driverbase.BuildGetObjectsRecordReader`.
- **`go/connection.go`**: Added fast-path call in `GetObjects()` before the existing SQL template path. Falls through when the direct path is not applicable.


go/get_objects_direct.go
```diff
@@ -0,0 +1,382 @@
+// Copyright (c) 2025 ADBC Drivers Contributors
```
lidavidm:
```suggestion
// Copyright (c) 2026 ADBC Drivers Contributors
```

go/get_objects_direct.go
```diff
@@ -0,0 +1,382 @@
...
+// execShowSchemas executes a SHOW TERSE SCHEMAS command and reads the results directly.
+func (c *connectionImpl) execShowSchemas(ctx context.Context, pattern *string, suffix string) (entries []schemaEntry, err error) {
+	query := "SHOW TERSE /* ADBC:getObjects */ " + objSchemas
+	query = addLike(query, pattern)
+	query += suffix
+
+	rows, err := c.cn.QueryContext(ctx, query, nil)
+	if err != nil {
+		var sfErr *gosnowflake.SnowflakeError
+		if errors.As(err, &sfErr) && sfErr.Number == 2043 {
```
lidavidm: What is error 2043?

go/get_objects_direct.go
```diff
@@ -0,0 +1,382 @@
...
+	cols := rows.Columns()
+	nameIdx, dbIdx := -1, -1
+	for i, col := range cols {
+		switch col {
+		case "name":
+			nameIdx = i
+		case "database_name":
+			dbIdx = i
+		}
+	}
```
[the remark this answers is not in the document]
lidavidm: I think that's useful to note in the code itself

lidavidm (approved):
LGTM, would be good to document that SHOW TERSE doesn't have a fixed column order though

frbvianna:
Hi @zeroshade, I'm updating to the latest ADBC and gosnowflake driver versions and see that the ADBC GetObjects call is taking much longer now at the table scope.
For any table, it used to take ~ 10 seconds at most, and now it's taking 120~150 seconds. I assume this may be behind it. Could you please confirm?

zeroshade:
@frbvianna I did a bit of digging into this:

- At ~100k database objects for GetObjects, the both the path before this and the new path were at parity in my testing (~25 - 30s)
- The cost was dominated by the server-side `SHOW TERSE OBJECTS` execution which is identical in both scenarios. 
- The client-side row read that was added here is showing <1s of runtime even at 100k rows (SHOW results come back largely inline)
- The direct path that was added uses fewer round-trips than the old path.

Could you share the following so I can try to reproduce the slowdown you're seeing?

- The exact `GetObjects` call you're making (depth, catalog/dbSchema/tableName/tableType params). In particular, does the catalog name contain any underscores? The new direct path only engages for a specific underscore-free catalog scenario, so that determines whether you're even hitting the new code.
- Rough object/schema counts in the depth you're querying GetObjects with
- The driver version you were using for the before and after tests (are you able to reproduce the faster runtime with an older version of the driver? and then immediately upgrade and show the slower runtime?)
- Can you go to the query history for your account and look at the execution time of the `SHOW TERSE ...` queries issued during the slow `GetObjects` call? (look for the `/* ADBC:getObjects */` tag). This way we can see how much of the time is server-side vs something client-side in the driver.

Thanks! Hopefully we can narrow down whether this is at fault or if something else is going on.

frbvianna:
Hi @zeroshade, thanks for looking into it. Replying below.

> The exact GetObjects call you're making (depth, catalog/dbSchema/tableName/tableType params). In particular, does the catalog name contain any underscores? The new direct path only engages for a specific underscore-free catalog scenario, so that determines whether you're even hitting the new code.

I've changed some names to avoid exposing sensitive info, but preserved their styling / formats (whether they include underscores or not):
 
- Object depth: adbc.ObjectDepthColumns
- Catalog: `DB_NAME`
- Schema: `TEST_SCHEMA`
- Table: `LINEITEM`

> Rough object/schema counts in the depth you're querying GetObjects with

The table (LINEITEM from TPC-H) has 16 columns, but I've tested with a table of only 1 column, and it still takes longer than a minute.

> The driver version you were using for the before and after tests (are you able to reproduce the faster runtime with an older version of the driver? and then immediately upgrade and show the slower runtime?)

We still used the latest version available for github.com/apache/arrow-adbc/go/adbc (v1.11.0), with github.com/snowflakedb/gosnowflake v1.19.1, and do not see the performance issue in that version.
Now, switching to github.com/adbc-drivers/snowflake/go (v1.11.0) and github.com/snowflakedb/gosnowflake/v2 (v2.1.0), we see that it takes much longer to fetch object metadata.

> Can you go to the query history for your account and look at the execution time of the SHOW TERSE ... queries issued during the slow GetObjects call? (look for the /* ADBC:getObjects */ tag). This way we can see how much of the time is server-side vs something client-side in the driver.

I executed again and it took around 92 seconds. Here are the SHOW TERSE ... queries, it looks like most of the time is taken in the SHOW COLUMNS statement:

```json
{
    "QUERY_ID": "01c57772-080c-6d11-0035-0e03070a7502",
    "QUERY": "SHOW TERSE /* ADBC:getObjects */ DATABASES LIKE 'DB_NAME' IN ACCOUNT",
    "START_TIME": "2026-07-03 11:26:47.109 -0700",
    "EXECUTION_STATUS": "SUCCESS",
    "ROWS_PRODUCED": "0",
    "TOTAL_ELAPSED_TIME": "179"
},
{
    "QUERY_ID": "01c57772-080c-7867-0035-0e03070a6362",
    "QUERY": "SHOW UNIQUE KEYS /* ADBC:getObjectsTables */ IN ACCOUNT",
    "START_TIME": "2026-07-03 11:26:47.106 -0700",
    "EXECUTION_STATUS": "SUCCESS",
    "ROWS_PRODUCED": "0",
    "TOTAL_ELAPSED_TIME": "406"
},
{
    "QUERY_ID": "01c57772-080c-6d11-0035-0e03070a7506",
    "QUERY": "SHOW IMPORTED KEYS /* ADBC:getObjectsTables */ IN ACCOUNT",
    "START_TIME": "2026-07-03 11:26:47.103 -0700",
    "EXECUTION_STATUS": "SUCCESS",
    "ROWS_PRODUCED": "0",
    "TOTAL_ELAPSED_TIME": "267"
},
{
    "QUERY_ID": "01c57772-080c-7eea-0035-0e03070a800e",
    "QUERY": "SHOW COLUMNS /* ADBC:getObjects */ IN ACCOUNT",
    "START_TIME": "2026-07-03 11:26:46.984 -0700",
    "EXECUTION_STATUS": "SUCCESS",
    "ROWS_PRODUCED": "0",
    "TOTAL_ELAPSED_TIME": "84866"
},
{
    "QUERY_ID": "01c57772-080c-7867-0035-0e03070a635a",
    "QUERY": "SHOW TERSE /* ADBC:getObjects */ SCHEMAS LIKE 'TEST_SCHEMA' IN ACCOUNT",
    "START_TIME": "2026-07-03 11:26:46.408 -0700",
    "EXECUTION_STATUS": "SUCCESS",
    "ROWS_PRODUCED": "0",
    "TOTAL_ELAPSED_TIME": "333"
}
```

frbvianna:
Hi @zeroshade, do you have any updates on this? Updating the driver is currently stalled for us due to this problem. Should we create a separate GitHub issue?

zeroshade:
My bad, I forgot to update here. I did a bunch of digging. I'll summarize it here and then create a new Issue to track it, TL;DR: the issue is actually introduced by #152, not by this one.

The *root* cause of the issue is actually the `SHOW COLUMNS /* ADBC:getObjects */ IN ACCOUNT` query. Looking at your timing ~85s of the ~92s was the server-side execution of this. The problem is the `IN ACCOUNT` part of the query. Because your catalog name has an underscore (`DB_NAME`), and the ADBC spec treats the arguments as `LIKE` patterns (`_` and `%` are wildcards), we have to assume the catalog *might* match more than one database, so the driver can't scope to a single DB and falls back to the `IN ACCOUNT`. This is also why the column count of your target table doesn't end up making a difference.

I did a test using different scopes on a test account:

| `SHOW COLUMNS` scope | elapsed |
|---|---|
| `IN ACCOUNT` | 18.4 s |
| `IN DATABASE` | 4.8 s |
| `IN SCHEMA` | 0.52 s |
| `IN TABLE` | 0.39 s |

~47x here, and much larger on a big account (your 85s).


This was introduced by #152 to read `byteLength` for BINARY column sizes. This PR only affects the DBSchemas/Tables depth, so it doesn't run the `ObjectDepthColumns` call, and wouldn't hit this change since you had an underscore in your catalog name. The reason the older version was faster was that it read column metadata straight from `information_schema.columns` and never issued a `SHOW COLUMNS` query at all.

To fix this, we can scope the `SHOW COLUMNS` query to the specific database/schema/table when concrete names are provided, rather than dropping to `IN ACCOUNT` just because a name contains an underscore. I'll put up a PR shortly and link it here. (If you need a stopgap and only require table listings, `GetObjects` at Tables depth avoids the `SHOW COLUMNS` path — but the real fix is the scoping.)

PR #114, opened 2026-04-16
fix(go): CALL <stored_proc> fails with arrow/ipc arrow due to EOF

### LINKED ISSUE #115: [GO] Calling stored procedures using CALL <stored_proc> throws arrow/ipc error
When calling a stored procedure such as CALL PUBLIC.GetNYCTaxiGreenData() the driver throws the error `[Snowflake] arrow/ipc: could not read message schema: arrow/ipc: could not read message metadata: unexpected EOF` because the data is returned as JSON data instead of the requested Arrow data.

davidhcoe:
## What's Changed

-  Stored procedures return result sets in JSON format, but ADBC cannot handle JSON batches (JSON-to-Arrow conversions) after the first iteration.
-  The procedures use a pattern (returns table without explicit column definitions) that doesn't work with Power BI's SELECT * FROM approach
-  Procedures contain DDL operations with temporary tables, which hit tabular SELECT table function limitations

- added `adbc.snowflake.sql.client_option.stream_retry_enabled` that, when true, enables buffers entire batches and retries on failure. Default is false to stay on the existing path.

Closes https://github.com/adbc-drivers/snowflake/issues/115


davidhcoe:
Evidently https://github.com/snowflakedb/gosnowflake/pull/1773 is happening in parallel so maybe we wait on that change to see if this is still necessary?

davidhcoe:
Anything more needed on this @lidavidm @zeroshade ?

go/record_reader.go
```diff
@@ -559,7 +561,198 @@ type reader struct {
...
+
+// readBatchRecords reads all Arrow records from a Snowflake batch with retries.
+// It does not buffer the raw byte stream; instead it reads Arrow IPC messages
+// directly from the network and accumulates transformed record batches in memory.
+// If a stream fails mid-read, all partial records are released and the entire
+// batch is re-downloaded. Records are only returned on full success, preventing
+// partial results from entering channels.
+func readBatchRecords(ctx context.Context, batch batchStreamer, alloc memory.Allocator, transform recordTransformer, maxRetries int) ([]arrow.RecordBatch, error) {
+	var lastErr error
+	for attempt := 0; attempt <= maxRetries; attempt++ {
```
lidavidm, line 596: nit: `for attempt := range maxRetries` (I need to fix the `go fix` check...)

lidavidm (approved):
Kicked off CI: https://github.com/adbc-drivers/snowflake/actions/runs/25413486408

PR #117, opened 2026-04-25
feat: GeoArrow GEOGRAPHY/GEOMETRY ingest, export, and EWKB CRS readback

jatorre:
## Summary

End-to-end GeoArrow support for `GEOGRAPHY` and `GEOMETRY` columns. **This PR consolidates and supersedes #99 (ingest) and #100 (export)** so reviewers have one place to look at the full geo story; both are now closed in favor of this one.

There are four pieces:

1. **Bulk ingest** (was #99). Detects `geoarrow.wkb` columns in the Arrow input and routes them through PUT → COPY INTO with an inline `TO_GEOGRAPHY`/`TO_GEOMETRY` transform.
2. **Per-column GEOMETRY auto-detect on ingest** (new). When the user hasn't pinned `adbc.snowflake.statement.ingest_geo_type`, each `geoarrow.wkb` column whose CRS is non-`EPSG:4326` is promoted to `GEOMETRY` so its SRID survives the round trip; sibling 4326 columns stay `GEOGRAPHY`. Lets clients pass `geoarrow.wkb` with `{"crs":"EPSG:3857"}` and get the right result without setting a statement option (which DuckDB's `adbc_insert` wrapper, for one, doesn't expose).
3. **Export** (was #100). Sets `GEOGRAPHY/GEOMETRY_OUTPUT_FORMAT` so values arrive as binary, and tags the result columns as `geoarrow.wkb`.
4. **CRS readback (new).** Switches the output format default to `EWKB` and lifts the SRID off the wire, so consumers see `geoarrow.wkb` with `{"crs":"EPSG:N"}` field metadata.

## How CRS readback works

- `database.go` defaults both `GEOGRAPHY_OUTPUT_FORMAT` and `GEOMETRY_OUTPUT_FORMAT` to `EWKB` so every value carries its SRID inline.
- `record_reader.go` peeks the first IPC record on each query, classifies any binary column whose values look like a valid OGC/EWKB geometry header as geo, and lifts the SRID into `geoarrow.wkb` field metadata as `{"crs":"EPSG:N"}`.
- A per-row column transformer strips the EWKB SRID prefix so consumers see standard ISO/OGC WKB on the Arrow wire (the field metadata still claims `geoarrow.wkb`, which is plain WKB by spec).
- No SQL-text parsing, no `DESCRIBE TABLE`, no extra `ST_SRID` round-trips.

## Why peek the bytes instead of reading the result schema

@lidavidm asked on #100 about the Snowflake REST API reporting column types as `"binary"` instead of `"geography"`/`"geometry"` when the WKB/EWKB output format is set — both `rowtype` and `DESCRIBE RESULT` are affected. (cc @sfc-gh-obielov, Snowflake side.)

Reading SRID off the wire makes the driver independent of that REST API behavior:

- Works for any query shape — joins, CTEs, `ST_Transform`, aliased computed columns. (Verified locally.)
- No reliance on rowtype/DESCRIBE RESULT correctness.
- Eliminates an earlier `FROM <table>` regex + `DESC TABLE` + `ST_SRID(col) GROUP BY 1 LIMIT 2` round-trip that only worked for single-table scans.
- Fixes a latent labeling issue: the previous path advertised `geoarrow.wkb` while the bytes were actually EWKB; geoarrow.wkb is plain WKB by spec.

If/when Snowflake fixes the rowtype metadata, the peek is still useful (covers per-row SRID for mixed-SRID columns and computed expressions), so this isn't blocked on a server-side change.

## Limitations

- **GEOMETRY columns with no SRID set** (Snowflake stores SRID 0): emit pure WKB without the EWKB SRID flag, so they arrive as plain `BINARY` without `geoarrow.wkb` tagging. They have no CRS to advertise either way.
- **Empty result sets**: cannot be peeked; their binary columns return as plain `BINARY`. Caller has zero rows to interpret.
- **Mixed-SRID columns** (multiple distinct SRIDs in the same column): tagged as `geoarrow.wkb` without a column-level CRS; per-row WKB is still correct.

## Test plan

- [x] `go test -tags assert -short ./...` in `go/` passes.
- [x] `go build -tags driverlib -buildmode=c-shared` produces a working dylib.
- [x] **End-to-end SRID survival via the actual driver import path**: DuckDB sends `geoarrow.wkb` with `{"crs":"EPSG:3857"}` to `adbc_insert`; the driver auto-detects GEOMETRY, applies `ST_SETSRID` inline during COPY INTO; readback returns `GEOMETRY('EPSG:3857')` on the DuckDB consumer side via the EWKB-peek path. No statement options set anywhere.
- [x] Verified arbitrary-query coverage that a SQL-parser approach would have missed: joins, CTEs, aliased computed columns. All produce correctly tagged geo columns with the right per-column CRS.
- [ ] Validation suite (`pixi run validate`) — not run locally; happy to wire it up in CI.

cc @lidavidm @zeroshade @sfc-gh-obielov


go/statement.go
```diff
@@ -367,14 +385,20 @@ func (st *statement) SetSqlQuery(query string) error {
...
 	switch dt.ID() {
 	case arrow.EXTENSION:
-		return toSnowflakeType(dt.(arrow.ExtensionType).StorageType())
+		ext := dt.(arrow.ExtensionType)
+		switch ext.ExtensionName() {
+		case "geoarrow.wkb", "geoarrow.wkb_view", "geoarrow.wkt", "geoarrow.wkt_view":
+			return geoType
+		default:
+			return toSnowflakeType(ext.StorageType(), geoType)
+		}
```
lidavidm: (This is what bothers me about extension type support in Arrow libraries) Because we don't actually implement the extension type/register it globally (and IMO we shouldn't), GeoArrow types won't be visible as an extension type per se, instead you have to check the field metadata on any binary types. The Go test, by implementing the extension type itself, masks this, but when we use the C Data Interface I believe this won't work.

go/statement_test.go
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) 2025 ADBC Drivers Contributors
```
lidavidm: (nit: it is 2026)

go/statement.go
```diff
@@ -493,16 +530,333 @@ func (st *statement) executeIngest(ctx context.Context) (int64, error) {
...
+//
+// Snowflake's COPY INTO from Parquet cannot load WKB directly into GEOGRAPHY/GEOMETRY
+// columns — only CSV and JSON/AVRO support direct geospatial loading from stages.
+// See: https://docs.snowflake.com/en/sql-reference/data-types-geospatial#loading-geospatial-data-from-stages
+//
+// We work around this with a CTAS pattern: rename the staging table, create the
+// final table with TO_GEOGRAPHY/TO_GEOMETRY conversion, then drop staging.
+//
+// TODO: Investigate using a COPY transform (SELECT ... FROM @stage) to convert
+// inline during COPY INTO, which would avoid the rename+CTAS overhead.
```
lidavidm: Don't we do this above? When do we need this?

go/statement.go
jatorre: Agreed — removed the geoarrow branch from `toSnowflakeType` entirely. Detection now lives in `buildCopyQuery`, which checks both `arrow.EXTENSION` and `ARROW:extension:name` field metadata, and feeds a geo-type override map into `initIngest`. Verified end-to-end against a real Snowflake account with a column delivered as plain `BINARY` + field metadata only (no registered extension type) — the table gets a `GEOGRAPHY` column as expected.

go/statement_test.go
jatorre: Fixed.

go/statement.go
jatorre: Right — `buildCopyQuery` uses identical detection logic, so `convertGeoColumns` was only reachable when there were no geo columns (in which case it was a no-op). Deleted, along with the `usedGeoTransform` plumbing.

go/record_reader.go
```diff
@@ -48,11 +49,187 @@ import (
...
+// codes (Point..GeometryCollection, optionally with Z/M/ZM extensions).
+func validGeoTypeCode(t uint32) bool {
+	switch {
+	case t >= 1 && t <= 7,
+		t >= 1001 && t <= 1007,
+		t >= 2001 && t <= 2007,
+		t >= 3001 && t <= 3007:
+		return true
+	}
+	return false
```
lidavidm, line 114: nit: just `return t >= 1 && t <= 7 || ...` directly?

lidavidm (approved):

lidavidm:
Triggering CI here: https://github.com/adbc-drivers/snowflake/pull/128

go/connection.go
```diff
@@ -509,9 +509,21 @@ func (c *connectionImpl) toArrowField(columnInfo driverbase.ColumnInfo) arrow.Fi
...
 		}
 	case "GEOGRAPHY":
-		fallthrough
+		// With GEOGRAPHY_OUTPUT_FORMAT=WKB, data arrives as binary WKB.
+		// GEOGRAPHY is always WGS84 (SRID 4326).
+		field.Type = arrow.BinaryTypes.Binary
+		field.Metadata = arrow.MetadataFrom(map[string]string{
+			"ARROW:extension:name":     "geoarrow.wkb",
+			"ARROW:extension:metadata": `{"crs":"EPSG:4326"}`,
+		})
```
paleolimbot, line 518: This won't be GEOGRAPHY unless the edge definition is set (e.g., pyarrow won't write it to Parquet as GEOGRAPHY unless this is the case):

```suggestion
		// With GEOGRAPHY_OUTPUT_FORMAT=WKB, data arrives as binary WKB.
		// GEOGRAPHY is always WGS84 (SRID 4326).
		field.Type = arrow.BinaryTypes.Binary
		field.Metadata = arrow.MetadataFrom(map[string]string{
			"ARROW:extension:name":     "geoarrow.wkb",
			"ARROW:extension:metadata": `{"crs":"EPSG:4326",edges:"spherical"}`,
		})
```

go/connection.go
```diff
@@ -554,9 +566,16 @@ func descToField(name, typ, isnull, primary string, comment sql.NullString, maxT
 		case "VARIANT":
 			field.Type = arrow.BinaryTypes.String
 		case "GEOGRAPHY":
-			fallthrough
+			field.Type = arrow.BinaryTypes.Binary
+			field.Metadata = arrow.MetadataFrom(map[string]string{
+				"ARROW:extension:name":     "geoarrow.wkb",
+				"ARROW:extension:metadata": `{"crs":"EPSG:4326"}`,
+			})
```
paleolimbot, line 573: Also here

go/connection.go
```diff
@@ -509,9 +509,21 @@ func (c *connectionImpl) toArrowField(columnInfo driverbase.ColumnInfo) arrow.Fi
...
+		field.Type = arrow.BinaryTypes.Binary
+		field.Metadata = arrow.MetadataFrom(map[string]string{
+			"ARROW:extension:name":     "geoarrow.wkb",
+			"ARROW:extension:metadata": `{"crs":"EPSG:4326"}`,
+		})
 	case "GEOMETRY":
-		field.Type = arrow.BinaryTypes.String
+		// With GEOMETRY_OUTPUT_FORMAT=WKB, data arrives as binary WKB.
+		// TODO: SRID for GEOMETRY requires inspecting data or a separate query.
+		// Same cross-driver issue as adbc-drivers/redshift#2 and adbc-drivers/databricks#350.
```
paleolimbot, line 522: It seems like this is implemented in this PR...can the TODO be removed?

go/record_reader.go
```diff
@@ -48,11 +49,187 @@ import (
...
+	return bldr.NewArray(), nil
+}
+
+// analyzeGeoFromBatch walks the binary columns of a record batch and returns
+// a map of column-name → geoColumnInfo for any column whose first non-null
+// value parses as an OGC/EWKB geometry header. The first SRID seen is taken
+// as the column-level SRID; columns whose subsequent rows disagree are still
+// tagged as geo but without column-level CRS metadata.
+func analyzeGeoFromBatch(rec arrow.RecordBatch) map[string]geoColumnInfo {
+	if rec == nil || rec.NumRows() == 0 {
```
paleolimbot, line 171: Cool!

go/statement.go
```diff
@@ -493,16 +529,205 @@ func (st *statement) executeIngest(ctx context.Context) (int64, error) {
...
+			ext := f.Type.(arrow.ExtensionType)
+			extName = ext.ExtensionName()
+			extMeta = ext.Serialize()
+		} else if name, ok := f.Metadata.GetValue("ARROW:extension:name"); ok {
+			extName = name
+			extMeta, _ = f.Metadata.GetValue("ARROW:extension:metadata")
+		}
+
+		switch extName {
+		case "geoarrow.wkb", "geoarrow.wkb_view", "geoarrow.wkt", "geoarrow.wkt_view":
```
paleolimbot, line 593:
```suggestion
		case "geoarrow.wkb", "geoarrow.wkt":
```

`"geoarrow.wkb_view"` and `"geoarrow.wkt_view"` aren't extension names (and for what it's worth I'm not sure anything actually uses `geoarrow.wkt`)

go/statement.go
```diff
@@ -493,16 +529,205 @@ func (st *statement) executeIngest(ctx context.Context) (int64, error) {
...
+func (opts *ingestOptions) resolveGeoType(extMeta string) string {
+	if opts.geoTypeExplicit {
+		return opts.geoType
+	}
+	srid := extractSRIDFromMeta(extMeta)
+	if srid != 0 && srid != 4326 {
+		return "geometry"
+	}
+	return "geography"
+}
```
paleolimbot, line 678: I don't think this is correct...geometry can have an SRID of 4326 (has planar edges); geography has spherical edges. You can check the `"edges"` key (if it exists and is `"spherical"` you can use geography).

go/statement_test.go
```diff
@@ -0,0 +1,202 @@
...
+func (g *geoArrowType) Serialize() string     { return g.meta }
+func (g *geoArrowType) Deserialize(storage arrow.DataType, data string) (arrow.ExtensionType, error) {
+	return newGeoArrowType(g.name, storage, data), nil
+}
+func (g *geoArrowType) ExtensionEquals(other arrow.ExtensionType) bool {
+	return g.ExtensionName() == other.ExtensionName()
+}
+func (g *geoArrowType) ArrayType() reflect.Type {
+	return reflect.TypeFor[array.ExtensionArrayBase]()
+}
```
paleolimbot, line 52: This has come up enough times that it's probably worth canonicalizing it in the near future so that it's auto-registered in more places

paleolimbot (commented):
Cool!

Feel free to ping me on geoarrow extension type stuff...always happy to grow the ecosystem!

go/statement_test.go
lidavidm: It's an artifact of tests (it should never come up through the DLL boundary)

go/statement.go
lidavidm: :facepalm: I can't believe I missed that

go/connection.go
lidavidm: Hmm, we'll need to go fix this in various places then. Should we expect/require the same on ingest too (i.e. refuse to use the database's GEOGRAPHY type unless this is set?)

paleolimbot: I'm not sure I quite followed how that option is set...a user opt-in to use the geography type is fine for any geo input (but `"edges:"spherical"` should be automatically mapped to geography and back).

go/statement_test.go
paleolimbot: It's a test artifact that's shown up many times now (at least: one in pyarrow, one in Arrow C++ for Parquet, one in GDAL that's not even for testing) but it hasn't seemed to cause issues (that anybody has told me about 🙂 )

go/connection.go
lidavidm: Just to be clear: should we map _any_ type with `edges:spherical` to geography? Or only if the SRID is 4326 *and* `edges:spherical`? (And when the user explicitly opts in to it)

lidavidm:
I filed https://github.com/adbc-drivers/snowflake/issues/140, I'll fix these when I get a chance - thanks Dewey for checking over this!

go/connection.go
paleolimbot: > Just to be clear: should we map any type with edges:spherical to geography?

Yes.

> Or only if the SRID is 4326 and edges:spherical?

I am not sure if you want to get into the validating-a-crs-for-geography, but if you do (I think there might be some code here already that does), Snowflake geographies MUST be EPSG:4326/OGC:CRS84 (or PROJJSON equivalent). Some other engines can handle any GeographicCRS or omitted CRS.

https://github.com/apache/sedona-db/blob/ea969b72bafc83f9b8869f7bafadec68977d933b/rust/sedona-functions/src/st_setsrid.rs#L544-L572 if you're curious.

> (And when the user explicitly opts in to it)

Opt in is a good escape hatch if the guessed type is wrong or the CRS validation is too strict.

CurtHagenlocher:
FWIW, this is a breaking change :(. Shouldn't this kind of thing be driven by a connection property?

lidavidm:
@CurtHagenlocher because the result type is potentially different?

The thing is, having a ton of flags is also not great...a single flag to disable extension types altogether may be fine though?

CurtHagenlocher:
> @CurtHagenlocher because the result type is potentially different?

Not potentially different; actually different :).

> The thing is, having a ton of flags is also not great...a single flag to disable extension types altogether may be fine though?

I agree that lots of flags is awful but added one anyway in #158.

lidavidm:
I filed https://github.com/adbc-drivers/docs.adbc-drivers.org/issues/94

PR #118, opened 2026-04-27
fix(go): linting

zeroshade:
Fix lint problem introduced in #105


lidavidm:
@zeroshade do you mind running the CI for changes?

PR #122, opened 2026-05-11
fix(go): number(38, 11) crashes when UseHighPrecision is false

alokdevrani:
Fix: NUMBER(38, 11) crash when use_high_precision=false
**Problem**
When querying a NUMBER(p, s) column with precision ≥ 19 and scale > 0 via the ADBC stack (Implementation="2.0") with use_high_precision=false, Snowflake sends the column as int64 on the Arrow wire (unscaled integer). The driver then attempted a direct int64 → float64 safe cast via compute.Divide, which Arrow rejects for values exceeding 2^53 (9,007,199,254,740,992):

[DataSource.Error] invalid: integer value 42135425651100000 not in range:-9007199254740992 to 9007199254740992
Root cause: In getTransformer (record_reader.go), the guard routing high-precision integers through the safe Decimal128 intermediate path was:

if srcMeta.Precision > 15 && srcMeta.Precision < 19  // ← missed precision ≥ 19

Any NUMBER(p, s) with p ≥ 19 and s > 0 fell through to the direct compute.Divide path, where unscaled values > 2^53 caused a range error.

**Fix**
Removed the incorrect upper bound in record_reader.go:

// Before
if srcMeta.Precision > 15 && srcMeta.Precision < 19
// After
if srcMeta.Precision > 15

Any NUMBER column with precision > 15 and scale > 0 now routes through the integerToDecimal128 intermediate path: int64 → Decimal128(20,0) (no range issue) → type metadata substituted to Decimal128(p,s) → UnsafeCast to float64. This avoids Arrow's safe-cast range check entirely.

**Test**
Added TestIntegerToDecimal128_LargeUnscaledValue in record_reader_test.go which directly verifies that a NUMBER(38, 11) unscaled value of 42135425651100000 (representing 421354.256511, exceeding 2^53) is correctly converted to float64 without error.

### DIFF
--- a/go/record_reader_test.go
+++ b/go/record_reader_test.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
-	"github.com/apache/arrow-go/v18/arrow/compute"
 	"github.com/apache/arrow-go/v18/arrow/ipc"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


go/record_reader_test.go
```diff
@@ -361,3 +362,52 @@ func TestReadBatchRecords_PartialRecordsReleasedOnRetry(t *testing.T) {
 	// partial records of the failed first attempt.
 	assert.EqualValues(t, 1, recs[0].NumRows())
 }
+
+// TestIntegerToDecimal128_LargeUnscaledValue is the regression test for the NUMBER(38, 11) crash.
```
lidavidm: I think this is mostly a test of arrow-go and not the driver? Also it doesn't even test the code path in question here?

alokdevrani: I refactored my code and created test for both paths.

davidhcoe:
@lidavidm , @zeroshade - fyi, Alok is from our team as well. Do you know why the Linux test is failing with 

FAILED validation/tests/test_query.py::TestQuery::test_get_table_schema[snowflake:10:type/select/int16] - adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: [snowflake] 260000: account is empty. Vendor code: 260000
FAILED validation/tests/test_query.py::TestQuery::test_get_table_schema[snowflake:10:type/select/int32] - adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: [snowflake] 260000: account is empty. Vendor code: 260000
FAILED validation/tests/test_query.py::TestQuery::test_get_table_schema[snowflake:10:type/select/int64] - adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: [snowflake] 260000: account is empty. Vendor code: 260000
FAILED validation/tests/test_query.py::TestQuery::test_get_table_schema[snowflake:10:type/select/string] - adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: [snowflake] 260000: account is empty. Vendor code: 260000

Is there something we need here that we are missing? Or is it a broader issue?

lidavidm:
All external PRs (e.g. from a fork) can't run the full CI and have to go through the manually-triggered pipeline (https://github.com/adbc-drivers/snowflake/actions/workflows/go_test_pr.yaml)

I should probably adjust the CI to not even try to run in this case or fail with something more explicit

lidavidm:
https://github.com/adbc-drivers/snowflake/actions/runs/26195078555

lidavidm (approved):

PR #123, opened 2026-05-20
fix: panic and crash on schema mismatch in Batch ops

pkit:
## What's Changed

Added a handler for panic in readRecords

Fixes https://github.com/apache/arrow-adbc/issues/2108

### DIFF
--- a/go/bulk_ingestion.go
+++ b/go/bulk_ingestion.go
@@ -333,8 +333,21 @@ func newWriterProps(mem memory.Allocator, opts *ingestOptions) (*parquet.WriterP
 	return parquetProps, arrowProps
 }
 
-func readRecords(ctx context.Context, rdr array.RecordReader, out chan<- arrow.RecordBatch) error {
+func readRecords(ctx context.Context, rdr array.RecordReader, out chan<- arrow.RecordBatch) (err error) {
 	defer close(out)
+	// The bound RecordReader is producer-supplied and may panic from
+	// Next(): arrow-go's C Data import, for example, panics when an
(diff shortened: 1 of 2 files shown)


go/bulk_ingestion.go
```diff
@@ -333,8 +333,21 @@ func newWriterProps(mem memory.Allocator, opts *ingestOptions) (*parquet.WriterP
...
 }
 
-func readRecords(ctx context.Context, rdr array.RecordReader, out chan<- arrow.RecordBatch) error {
+func readRecords(ctx context.Context, rdr array.RecordReader, out chan<- arrow.RecordBatch) (err error) {
 	defer close(out)
+	// The bound RecordReader is producer-supplied and may panic from
+	// Next(): arrow-go's C Data import, for example, panics when an
+	// inbound batch's column count does not match the advertised schema.
+	// Convert any such panic into an ADBC error so the caller gets a
+	// diagnosable failure instead of an aborted host process.
```
lidavidm, line 342: @zeroshade maybe this could be changed to not a panic upstream? In the meantime I'll add this case to the backlog of things we need to validate

pkit: The repro was there in https://github.com/apache/arrow-adbc/issues/2108 from 2024 
And I have a repro in this PR as a test.
I don't think any more validation is needed.

lidavidm: I mean adding this case to the cross-driver validation suite to catch this kind of bug in the future.

lidavidm (approved):

Issue #129, opened 2026-05-22
[Go] NUMBER(n, 0) values get truncated with use_high_precision=false


amoeba:
_This issue was originally filed as https://github.com/apache/arrow-adbc/issues/1277 by @CurtHagenlocher and has been duplicated here._

When the "high precision" switch is disabled, the Go Snowflake drivers return NUMBER(n, 0) values as int64s. For values which are too large to fit in 64 bits, it simply discards the upper 64 bits and returns the lower 64 bits without warning.

The "right thing" is particularly hard to figure out here because of the property of Snowflake that integer types are all aliases for NUMBER(38, 0), which means that even if the user thought they were creating an int32 columns it's actually stored as a decimal128-equivalent in the Snowflake schema. The best choice might be to error out when there's a value that's too large.

PR #131, opened 2026-05-22
test(go): add select validation tests for number

amoeba:
## What's Changed

I saw the issue https://github.com/adbc-drivers/snowflake/issues/129 and noticed the type mappings table doesn't mention this so I'm adding validation tests. In theory, this will cause the `NUMERIC` row in the Snowflake to Arrow table to change from jus showing `double` to showing `double, int64` with a footnote about the truncation.


lidavidm (approved):
Could also set the respective sql-type-names to `NUMERIC (use_high_precision = [true/false])` and then there will be two rows to clarify the behavior

lidavidm:
CI: https://github.com/adbc-drivers/snowflake/actions/runs/26374455330

lidavidm:
Hmm, it somehow came back as `int64`?

```
>       assert expected.type == actual.type, (
            f"Field types do not match: {path}{expected.name} ({expected.type}) != {path}{actual.name} ({actual.type})"
        )
E       AssertionError: Field types do not match: res (decimal128(38, 0)) != res (int64)
E       assert Decimal128Type(decimal128(38, 0)) == DataType(int64)
E        +  where Decimal128Type(decimal128(38, 0)) = pyarrow.Field<res: decimal128(38, 0)>.type
E        +  and   DataType(int64) = pyarrow.Field<res: int64>.type
```

lidavidm:
I'm going to fix this up.

lidavidm:
Superseded by https://github.com/adbc-drivers/snowflake/pull/142

PR #134, opened 2026-05-25
fix(go): preserve reader errors and reject truncated streams

### LINKED ISSUE #133: [Go] Snowflake Arrow stream reader can silently truncate results and leak goroutines on cancel
## What's happening

When consuming an Arrow result stream from the Snowflake Go driver, the
record reader has two related failure modes:

1. **Silent truncation.** If a chunk download ends prematurely (e.g. a
   mid-stream TCP RST from S3/Azure Blob, or a short read that the IPC
   layer happens to interpret as a clean end-of-stream), the reader's
   producer goroutine returns `rr.Err()` (which can be `nil` in that
   case), `group.Wait()` returns `nil`, and the consumer sees
   `Next() == false` with `Err() == nil`. The result set is incomplete,
   but no error is reported to the caller.

2. **Cancellation goroutine leak / late-cancel false error.** Two
   related variants observed:

   - When the consumer cancels mid-stream, the streaming code paths can
     return `nil` (from `rr.Err()`) instead of `ctx.Err()`, so the
     reader presents a "successful" partial result to the caller.
   - When cancellation lands during the trailing `rr.Next()` completion
     probe — after every expected row has already been emitted — the
     current code surfaces it as a generic error even though the batch
     completed successfully.

   In both variants, `concatReader.Next()` also drops a terminal error
   from the inner reader when stepping to the next reader, masking the
   real cause.

## Context

The underlying gosnowflake report is
[snowflakedb/gosnowflake#1789](https://github.com/snowflakedb/gosnowflake/issues/1789)
(SNOW-3479780). Two upstream fixes already landed there:

- [#1792](https://github.com/snowflakedb/gosnowflake/pull/1792) — adds
  `ArrowStreamBatch.Reset()` (mirror of the original
  [#1782](https://github.com/snowflakedb/gosnowflake/pull/1782) from
  @davidhcoe), which makes mid-stream retries possible.
- [#1795](https://github.com/snowflakedb/gosnowflake/pull/1795) — wraps
  opened Arrow chunk streams in a cancel-aware reader, so cancellation
  unblocks an in-flight body read and normalizes the close error to
  `ctx.Err()`.

Those address the gosnowflake-side symptoms, but the ADBC consumer
still needs to:

- Validate per-batch and per-result-set row counts so a short stream
  becomes a real error rather than a silent EOF.
- Preserve `rr.Err()` / `ctx.Err()` in every producer code path.
- Treat the "all rows already emitted, cancellation landed during the
  trailing completion probe" case as success.
- Preserve a failing inner reader's error in `concatReader.Next()`.

## Scope

This issue tracks only the ADBC-side fix. The companion gosnowflake
fixes are already merged on `master` and will be picked up on the next
dependency bump.

sfc-gh-boler:
## What's Changed

Fixes two related failure modes in the Snowflake Arrow stream record
reader.

**Silent truncation.** A chunk download that ends prematurely
(mid-stream TCP RST from S3/Azure Blob, or a short read that the IPC
layer interprets as a clean end-of-stream) used to surface as
`Next() == false, Err() == nil`, hiding an incomplete result set. The
streaming, retrying, and JSON paths now share a `finalizeBatchRead`
helper that compares actual rows against `b.NumRows()` (per batch) and
`ld.TotalRows()` (per result set), so a short stream becomes an
`adbc.StatusInvalidData` error instead of a silent EOF.

**Cancellation correctness.** The producer goroutines now preserve
`rr.Err()` / `ctx.Err()` in every path, the reader latches the first
producer error before advancing channels (so later prefetched rows
don't leak after an earlier batch fails), and a late cancellation that
lands during the trailing `rr.Next()` completion probe — after every
expected row was emitted — is treated as success rather than a generic
error. `concatReader.Next()` also preserves a failing inner reader's
error instead of dropping it while stepping forward.

Companion upstream fixes already merged in gosnowflake:

- [#1792](https://github.com/snowflakedb/gosnowflake/pull/1792) — adds
  `ArrowStreamBatch.Reset()` (mirrors @davidhcoe's
  [#1782](https://github.com/snowflakedb/gosnowflake/pull/1782)).
- [#1795](https://github.com/snowflakedb/gosnowflake/pull/1795) —
  cancel-aware reader so an in-flight body read unblocks promptly and
  normalizes the close error to `ctx.Err()`.

This PR is independent of those; the next gosnowflake dependency bump
will be a separate change.

### Tests

`go test -tags assert -race -count=1 ./...` plus
`pre-commit run --all-files` are clean. New tests cover:

- `TestReadBatchRecords_RetriesAfterRowCountMismatch`
- `TestTryReadBatch_TruncatedStreamFailsRowCountValidation`
- `TestStreamBatchToChannel_TruncatedStreamFailsRowCountValidation`
- `TestStreamBatchToChannel_CancellationReturnsContextError`
- `TestReaderLateCancellationAfterLastRecordReturnsSuccess`
- `TestReaderStopsAfterEarlierBatchFailureDespiteLaterPrefetch`
- `TestReaderCancellationSetsErrBeforeNextAndReleaseReturns`
- `TestConcatReaderPreservesInnerReaderError`
- `TestValidateRowCount_UsesNeutralMismatchMessage`

Closes #133.

### DIFF
--- a/go/concat_reader.go
+++ b/go/concat_reader.go
@@ -97,6 +97,10 @@ func (r *concatReader) Schema() *arrow.Schema {
 }
 func (r *concatReader) Next() bool {
 	for r.currentReader != nil && !r.currentReader.Next() {
+		if err := r.currentReader.Err(); err != nil {
+			r.err = err
+			break
+		}
 		r.nextReader()
 	}
 	if r.currentReader == nil || r.err != nil {
--- a/go/concat_reader_test.go
+++ b/go/concat_reader_test.go
@@ -1,0 +1,105 @@
+//
+// Copyright (c) 2025 ADBC Drivers Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package snowflake
+
+import (
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type mockReaderIter struct {
+	readers   []array.RecordReader
+	nextCalls int
+}
+
+func (m *mockReaderIter) Release() {}
+
+func (m *mockReaderIter) Next() (array.RecordReader, error) {
+	m.nextCalls++
+	if len(m.readers) == 0 {
+		return nil, io.EOF
+	}
+	reader := m.readers[0]
+	m.readers = m.readers[1:]
+	return reader, nil
+}
+
+type mockRecordReader struct {
+	schema *arrow.Schema
+	next   []bool
+	err    error
+}
+
+func (m *mockRecordReader) Release() {}
+
+func (m *mockRecordReader) Retain() {}
+
+func (m *mockRecordReader) Schema() *arrow.Schema {
+	return m.schema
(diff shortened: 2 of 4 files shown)


lidavidm:
CI: https://github.com/adbc-drivers/snowflake/actions/runs/26423999652

PR #136, opened 2026-05-26
fix(go): rescale scale-9 timestamps in microseconds mode (#135)

### LINKED ISSUE #135: [Go] Overflow when adbc.snowflake.sql.client_option.max_timestamp_precision = "microseconds"
Lines 514-516 of record_reader.go have
```
	if maxTimestampPrecision == Microseconds && originalArrowUnit == arrow.Nanosecond {
		return arrow.TimestampFromTime(time.UnixMicro(val.UnixMicro()), arrow.Microsecond)
	}
```
For values outside the range val.Unix() ∈ [-9.22e12, +9.22e12], this can produce an overflow. In the triggering example, the nanoseconds value was -6285681744183232512 (about 2/3 of int64.minValue) and after going through this logic produced 1506347332300000000 (same scale, wrong sign).

CurtHagenlocher:
Fixes #135.

NOTE: As per my comment in the issue, it's still not clear exactly what's going on here. The Snowflake UI shows a timestamp of `49704-03-31 14:31:40.000` for this value. I'm trying to figure out how it was created.

## Problem

When `adbc.snowflake.sql.client_option.max_timestamp_precision = "microseconds"`, a scale-9 (nanosecond) `TIMESTAMP_NTZ`/`TIMESTAMP_LTZ` value could come back with a wildly wrong date — e.g. a 2017-09-25 timestamp surfacing as year 49704.

The issue originally pointed at the `time.UnixMicro(val.UnixMicro())` overflow in `getArrowTimestampFromTime`, but that turned out to be a red herring. The two numbers in the report are the **same instant**: `1506347332300000000` µs is year 49704, and that instant's nanosecond representation overflows int64 to exactly `-6285681744183232512`. The microsecond conversion wasn't overflowing — the `val` it received was already inflated 1000×.

## Root cause

For a scale-9 column, `getArrowTimeUnit(9, Microseconds)` makes the field type `timestamp[µs]`, but the column's native wire unit is nanoseconds. The struct decoding path rescales correctly (via `time.Time`/`UnixMicro`), but the **single-`int64` fast path** used `compute.CastArray`, which *reinterprets* the raw nanosecond integer as microseconds instead of dividing by 1000:

```
int64(1506347332300000000 ns)  --cast-->  timestamp[µs] = 1506347332300000000  // year 49704
```

This only affects scale 9 + microseconds mode; every other scale/mode has matching native and field units, so the reinterpret was already correct.

## Fix

Add `castIntegerToTimestamp`, used by both non-struct paths:

- When the native unit equals the field unit (every default-mode case), reinterpret the integer as before — no behavior change.
- When they differ (ns → µs), rescale by integer division, **flooring toward negative infinity** to match the struct path's `UnixMicro` semantics (arrow's cast truncates toward zero, which would disagree by one unit for negative values carrying sub-microsecond precision).

A value of `-6285681744183232512` ns now yields `-6285681744183233` µs — no error, no overflow.

## Tests

`TestCastIntegerToTimestamp` covers:
- scale-9 microseconds rescale (the regression — year 2017, not 49704)
- default nanoseconds reinterpret (unchanged)
- timezone preservation (LTZ)
- null preservation
- negative value floors to `-6285681744183233` (not `-6285681744183232`)

`go build`, `go vet`, and the unit suite pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go/record_reader.go
+++ b/go/record_reader.go
@@ -354,7 +354,7 @@ func getTransformer(sc *arrow.Schema, ld gosnowflake.ArrowStreamLoader, useHighP
 			transformers[i] = func(ctx context.Context, a arrow.Array) (arrow.Array, error) {
 
 				if a.DataType().ID() != arrow.STRUCT {
-					return compute.CastArray(ctx, a, compute.SafeCastOptions(dt))
+					return castIntegerToTimestamp(ctx, a, dt, originalArrowUnit)
 				}
 
 				pool := compute.GetAllocator(ctx)
@@ -406,14 +406,7 @@ func getTransformer(sc *arrow.Schema, ld gosnowflake.ArrowStreamLoader, useHighP
 						tb.Append(v)
 					}
 				} else {
-					for i, t := range a.(*array.Int64).Int64Values() {
-						if a.IsNull(i) {
-							tb.AppendNull()
-							continue
-						}
-
-						tb.Append(arrow.Timestamp(t))
-					}
+					return castIntegerToTimestamp(ctx, a, dt, originalArrowUnit)
 				}
 				return tb.NewArray(), nil
 			}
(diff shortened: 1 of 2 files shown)


CurtHagenlocher:
Closing this — the corrected diagnosis is that it fixes a code path that isn't actually reached for well-formed data.

**Why the original framing was wrong.** Issue #135's symptom (microseconds mode returning a year-49704 value) turned out to be the read path *faithfully returning already-corrupt stored data*, not the read inflating a good value. The bad value originated upstream (a bad write), so reading it back — even correctly — shows year 49704.

**Why the patched path is unreachable.** This PR fixed the single-`int64` fast path's `int64(ns) → timestamp[µs]` cast, which does reinterpret rather than rescale. But a scale-9 column spans years 1–9999, which as nanoseconds is ~±2.5e20 and cannot fit in int64 (max ~9.2e18). So Snowflake must use the struct `{epoch_seconds, fraction}` encoding for scale-9 columns, and the driver's struct path already converts correctly via `getArrowTimestampFromTime`/`UnixMicro()`. `TestTimestampPrecision` corroborates this: it round-trips `0001-01-01` and `9999-12-31` at scale 9, which only the struct path can represent. The single-`int64` cast path is therefore not exercised for scale-9, which is the only scale where the microseconds option changes the unit.

In short: the microseconds option already handles real scale-9 data correctly, and the inflation reproduced here only occurs for a delivery form that doesn't happen for well-formed scale-9 data. Closing as not needed. If a future Snowflake encoding change ever delivers scale-9 as a single int64, this can be revisited as defensive hardening.

PR #142, opened 2026-06-02
test(go): add tests for NUMBER and highprecision option

lidavidm:


lidavidm:
@amoeba it seems the validation suite config here bitrotted so I took the liberty of taking your PR and fixing the suite config and upgrading various things in the process

PR #144, opened 2026-06-02
fix(go): address issues in geo implementation

lidavidm:


lidavidm:
@paleolimbot does this look more reasonable?

go/statement.go
```diff
@@ -710,37 +712,38 @@ func extractSRIDFromMeta(metadata string) int {
...
 	if err := json.Unmarshal(meta.CRS, &crsStr); err == nil {
 		if strings.HasPrefix(crsStr, "EPSG:") {
 			if code, err := strconv.Atoi(crsStr[5:]); err == nil {
-				return code
+				return code, meta.Edges
 			}
 		}
-		return 0
+		return 0, meta.Edges
 	}
```
paleolimbot, line 763: The other one that will show up here that is identical to `"EPSG:4326"` is `"OGC:CRS84"` (I wish this weren't true for what it's worth 😬 ). You can return an SRID of `4326` for this case.

go/statement.go
```diff
@@ -710,37 +712,38 @@ func extractSRIDFromMeta(metadata string) int {
...
 
 	var crs projCRS
 	if err := json.Unmarshal(meta.CRS, &crs); err == nil {
 		if strings.EqualFold(crs.ID.Authority, "EPSG") && crs.ID.Code != 0 {
-			return crs.ID.Code
+			return crs.ID.Code, meta.Edges
 		}
 	}
```
paleolimbot, line 770: The other one that will show up here is `{"authority":"OGC","code":"CRS84"}`. In Arrow C++ we also handle `{"authority":"EPSG","code":"4326"}` (where 4326 is a string) but I've never seen that in the wild.

go/statement.go
```diff
@@ -683,23 +683,25 @@ func (opts *ingestOptions) resolveGeoType(extMeta string) string {
...
-	srid := extractSRIDFromMeta(extMeta)
-	if srid != 0 && srid != 4326 {
-		return "geometry"
+	srid, edges := extractSRIDFromMeta(extMeta)
+	if srid == 4326 && edges == "spherical" {
+		return "geography"
 	}
-	return "geography"
+	return "geometry"
```
paleolimbot: I am a geospatial snob and so I probably would do:

```go
    if srid == 4326 && edges == "spherical" {
		return "geography"
	} else if srid == 4326 && edges == "spherical" {
        // error: force user-specified spatial type from the option because snowflake geography
        // only supports SRID 4326, so both geography and geometry are incorrect in different ways
    } else {
        return "geometry"
    }
```

...but your way is fine for all practical purposes (as long as you handle OGC:CRS84 below).

paleolimbot (approved):
Thank you! This is great...I am picky about these things but this is much better and I can file a follow up if you'd prefer to move on 🙂

go/statement.go
lidavidm: I adjusted this and added more tests!

paleolimbot (approved):
Thank you!

PR #154, opened 2026-06-12
feat(go): support ingesting LargeList columns

### LINKED ISSUE #126: [Go] Add support for ingesting List and LargeList
This feature was previously filed as https://github.com/apache/arrow-adbc/issues/1260 and has been duplicated here.

The Go Snowflake ADBC driver currently doesn't support ingesting List or LargeList fields. Trying to ingest the following table,

```python
import pyarrow as pa

schema = pa.schema(
    [
        pa.field("ints", pa.int64()),
        pa.field("strs", pa.string()),
        pa.field("lists", pa.large_list(pa.int64())), # or pa.field("lists", pa.list_(pa.int64())),
    ],
)

tbl = pa.table([[1, 2], ["a", None], [[1, 2, 3], None]], schema=schema)
tbl
# pyarrow.Table
# ints: int64
# strs: string
# lists: list<item: int64>
#   child 0, item: int64
# ----
# ints: [[1,2]]
# strs: [["a",null]]
# lists: [[[1,2,3],null]]
```

Results in:

> InternalError: INTERNAL: 002023 (22000): SQL compilation error:
> Expression type does not match column data type, expecting ARRAY but got VARCHAR(7) for column lists.

zeroshade:
## Summary

Adds support for ingesting Arrow `LargeList` columns (issue #126). Arrow `List`
columns already ingested into Snowflake `ARRAY` (mapped in `toSnowflakeType`,
read back as `arrow.json`), but `LargeList` failed at the Parquet write step
because arrow-go could not encode `LARGE_LIST` (apache/arrow-go#834). That bug
was fixed upstream by apache/arrow-go#838.

- Bump `github.com/apache/arrow-go/v18` from `v18.6.0` to the commit containing
  apache/arrow-go#838 (`v18.6.1-0.20260610195748-ef9e9859a205`). No released
  `v18.x` contains the fix yet, so this pins the merge commit; it can move to a
  tagged release once one ships. The only transitive change is
  `golang.org/x/sys` `v0.45.0` -> `v0.46.0`.
- Rewrite `TestParquetLargeList` to assert that a `LargeList` column now writes
  to Parquet and round-trips (it previously asserted the upstream breakage).
- Add `TestSqlIngestLargeListType`, an end-to-end ingestion test mirroring the
  existing `TestSqlIngestListType`, covering the original bug report that used
  `large_list`.

No driver logic changed: `toSnowflakeType` already maps `LIST` / `LARGE_LIST` /
`FIXED_SIZE_LIST` to `ARRAY`, and the `COPY INTO ... MATCH_BY_COLUMN_NAME` path
already loads Parquet lists into `ARRAY`. The only blocker was the arrow-go
writer, which the bump resolves.

## Testing

From `go/`: `go build ./...`, `go vet ./...`, `go test -tags assert ./...`
(`TestParquetLargeList` passes; integration tests skip without `SNOWFLAKE_URI`),
and pre-commit (golangci-lint, codespell, RAT) all pass.

## Note

`TestSqlIngestLargeListType` requires a live Snowflake instance and runs in CI.
Each list holds a single element, matching `TestSqlIngestListType`, because
Snowflake currently errors on multi-element repetition levels (pre-existing,
tracked separately).

Closes #126


go/bulk_ingestion_test.go
```diff
@@ -249,7 +250,6 @@ func makeRec(mem memory.Allocator, nCols, nRows int) arrow.RecordBatch {
 }
 
 func TestParquetLargeList(t *testing.T) {
```
lidavidm: I'd rather remove this test...the point was to be a canary if upstream was updated

go/driver_test.go
```diff
@@ -1421,6 +1421,91 @@ func (suite *SnowflakeTests) TestSqlIngestListType() {
 	suite.Require().NoError(rdr.Err())
 }
 
+func (suite *SnowflakeTests) TestSqlIngestLargeListType() {
```
lidavidm: Is this strictly necessary since it just duplicates the validation suite?

go/bulk_ingestion_test.go
zeroshade: Done in 84e2ee0 — removed `TestParquetLargeList`. Agreed: the canary has served its purpose now that the arrow-go LARGE_LIST Parquet writer fix (apache/arrow-go#838) has landed.

go/driver_test.go
zeroshade: Done in 84e2ee0 — removed `TestSqlIngestLargeListType`. You're right that it duplicates the validation suite: `large_list.txtcase` already exercises this ingestion path (and more thoroughly, with nested structs, nulls, and empty lists), so the hand-written test was redundant.

zeroshade:
Superseded by #160 (merged to main as d7cf709), which addresses #126 with byte-identical changes: same arrow-go bump to v18.6.1-0.20260610195748-ef9e9859a205, same removal of the TestParquetLargeList xfail, and same large_list.txtcase update. main already contains everything in this branch, so the net diff here is empty. Closing as superseded.

PR #158, opened 2026-06-17
feat: allow previous behavior for geo types to be opted into

CurtHagenlocher:
## What's Changed

#117 broke the behavior for geometry and geography types by changing the type of the returned data. This change adds a flag that allows the old behavior to be opted back into.

### DIFF
--- a/go/driver.go
+++ b/go/driver.go
@@ -99,6 +99,18 @@ const (
 	// original inline streaming path. Default is disabled.
 	OptionStreamRetryEnabled = "adbc.snowflake.sql.client_option.stream_retry_enabled"
 
+	// OptionGeographyOutputFormat controls the output format for GEOGRAPHY
+	// columns. Valid values are "EWKB" and "GeoJSON" (default, case-insensitive).
+	// When set to EWKB, GEOGRAPHY columns are returned as geoarrow.wkb binary
+	// with CRS metadata. When set to GeoJSON, they are returned as UTF-8 text.
+	OptionGeographyOutputFormat = "adbc.snowflake.sql.client_option.geography_output_format"
+
+	// OptionGeometryOutputFormat controls the output format for GEOMETRY
+	// columns. Valid values are "EWKB" and "GeoJSON" (default, case-insensitive).
+	// When set to EWKB, GEOMETRY columns are returned as geoarrow.wkb binary
+	// with CRS metadata. When set to GeoJSON, they are returned as UTF-8 text.
+	OptionGeometryOutputFormat = "adbc.snowflake.sql.client_option.geometry_output_format"
+
 	OptionApplicationName  = "adbc.snowflake.sql.client_option.app_name"
 	OptionSSLSkipVerify    = "adbc.snowflake.sql.client_option.tls_skip_verify"
 	OptionOCSPFailOpenMode = "adbc.snowflake.sql.client_option.ocsp_fail_open_mode"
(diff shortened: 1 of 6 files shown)


go/driver.go
```diff
@@ -99,6 +99,12 @@ const (
 	// original inline streaming path. Default is disabled.
 	OptionStreamRetryEnabled = "adbc.snowflake.sql.client_option.stream_retry_enabled"
 
+	// OptionGeometryAsText controls how GEOGRAPHY and GEOMETRY columns are
+	// returned. When enabled, these columns are returned as UTF-8 text
+	// (the pre-GeoArrow behavior). When disabled (the default), they are
+	// returned as geoarrow.wkb binary with CRS metadata.
+	OptionGeometryAsText = "adbc.snowflake.sql.client_option.geometry_as_text"
```
CurtHagenlocher: I know the name sucks. Feel free to suggest another.

lidavidm (commented):
How about having options `snowflake.session.geography_output_format`/`snowflake.session.geometry_output_format` (do we already have options for other session parameters that we need to follow?) that defaults to `GeoJSON` and can be set to `EWKB`? (With the other options currently being unsupported)

CurtHagenlocher:
> How about having options `snowflake.session.geography_output_format`/`snowflake.session.geometry_output_format` (do we already have options for other session parameters that we need to follow?) that defaults to `GeoJSON` and can be set to `EWKB`? (With the other options currently being unsupported)

That's fine with me if you think it makes sense.

I think the longer-term approach would be to more generically let the consumer specify what Arrow type they want to bind each column to. (EDIT: Oh yeah, `AdbcStatementRequestSchema`)

EDIT: is it possible that consumers have already taken a dependency on the changed behavior? I saw some links from that PR to other repositories.

lidavidm:
> That's fine with me if you think it makes sense.

Looks like we have `adbc.snowflake.sql.client_option.disable_telemetry` so I guess it should be `adbc.snowflake.sql.client_option.geometry_output_format`

> I think the longer-term approach would be to more generically let the consumer specify what Arrow type they want to bind each column to. (EDIT: Oh yeah, `AdbcStatementRequestSchema`)

Everything is held up on the distastefulness of serialized IPC schemas in metadata result tables ><

> EDIT: is it possible that consumers have already taken a dependency on the changed behavior? I saw some links from that PR to other repositories.

That PR isn't actually released, though, so there shouldn't be.
But let's let @jatorre chime in.

CurtHagenlocher:
Okay, I've made the changes you described @lidavidm.

> Everything is held up on the distastefulness of serialized IPC schemas in metadata result tables ><

Well, we could try again to land on a consensus JSON representation ;). Or a Parquet-variantified version thereof :P.

lidavidm (approved):
LGTM. I'll update any validation tests/docs etc. before merging and re-prepare the release

lidavidm:
I need to work out a better workflow for these tests.

https://github.com/adbc-drivers/snowflake/actions/runs/27725298705

CurtHagenlocher:
> ❌ **Test failed:** [CurtHagenlocher/snowflake@3f1e188](https://github.com/CurtHagenlocher/snowflake/commit/3f1e188c7460da8de6f6d94ffb0a028e9a1c0200) Workflow run:

Hard to be certain but it doesn't look like the change was responsible for those errors.

lidavidm:
Yeah, that's a flake.

lidavidm:
....I guess we never added validation tests for this

lidavidm:
I'll merge this, clean that up, and then tag the release

PR #161, opened 2026-06-19
fix(go): error on NUMBER values that overflow int64 with use_high_precision=false

### LINKED ISSUE #129: [Go] NUMBER(n, 0) values get truncated with use_high_precision=false

zeroshade:
## Summary

With `use_high_precision=false`, Snowflake `NUMBER(n,0)` columns are returned as Arrow `int64`. Because Snowflake stores integer types as `NUMBER(38,0)` (decimal128-equivalent), high-precision columns arrive as a `Decimal128` array, and the previous `compute.UnsafeCastOptions` cast **silently discarded the upper bits** for values that do not fit in int64 — returning a corrupted low-64-bit result with no error or warning.

Closes #129.

## Changes

- **`record_reader.go`** — scale-0 `Decimal128`/`Decimal256` columns are now converted via a new `decimalScale0ToInt64` transformer that range-checks each value and returns an `adbc.StatusInvalidData` error when it falls outside `[MinInt64, MaxInt64]`, instead of truncating. The bounds are checked manually (not via `compute.SafeCastOptions`) because arrow-go's safe decimal→int kernel tests `value >= MaxInt64` and therefore rejects `MaxInt64` itself.
- **`record_reader.go`** — preserve a transformer's `adbc.Error` status code through the record-transform loop, so the overflow surfaces as `InvalidData` rather than being re-wrapped as `Internal`.
- **Tests** — new `TestDecimalScale0ToInt64` unit test covering in-range values, the int64 min/max boundaries, and overflow; updated `TestIntDecimalLowPrecision` to expect a `StatusInvalidData` reader error for out-of-range `NUMBER(p,0)` values (precision ≥ 19).
- **Validation** — updated the `numeric_low_precision` caveat to document the new error-on-overflow behavior (it previously documented the silent truncation).

## Behavior change

Queries that previously returned a silently-truncated value for out-of-range `NUMBER` now return an error. Users who need the full value can set `use_high_precision=true` to read the column losslessly as `decimal128`.

## Testing

- `go test ./...` passes (unit tests; Snowflake integration tests skip without credentials).
- `go vet ./...` clean.

### DIFF
--- a/go/record_reader.go
+++ b/go/record_reader.go
@@ -42,6 +42,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/compute"
+	"github.com/apache/arrow-go/v18/arrow/decimal"
 	"github.com/apache/arrow-go/v18/arrow/ipc"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/snowflakedb/gosnowflake/v2"
@@ -259,6 +260,10 @@ func getRecTransformer(sc *arrow.Schema, tr []colTransformer) recordTransformer 
 
 		for i, col := range r.Columns() {
 			if cols[i], err = tr[i](ctx, col); err != nil {
+				var adbcErr adbc.Error
+				if errors.As(err, &adbcErr) {
+					return nil, adbcErr
+				}
 				return nil, errToAdbcErr(adbc.StatusInternal, err)
 			}
 		}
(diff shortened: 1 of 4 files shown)


go/record_reader.go
```diff
@@ -565,6 +571,67 @@ func scaledIntToFloat64Transformer(precision, scale int32) colTransformer {
 	}
 }
 
+// decimalScale0ToInt64 converts a scale-0 Decimal128/Decimal256 column to int64
+// for use_high_precision=false, returning a StatusInvalidData error for any value
+// outside the int64 range instead of silently truncating it (issue #129).
+//
+// The range is checked manually rather than with compute.SafeCastOptions because
+// arrow-go's safe decimal->int kernel tests `value >= MaxInt64` and so rejects
+// math.MaxInt64 itself; a manual check accepts both int64 bounds inclusively.
```
lidavidm, line 580: Er, should we fix this upstream?

zeroshade: good point. don't know why i didn't just do that when i saw this...

PR #162, opened 2026-06-19
test(go): expand geo validation coverage

### LINKED ISSUE #143: [Go] Add validation tests for geo implementation
These can be copied from one of the other drivers that implements it already, I think (minus adjustments for specific things for this driver like checking for `edges:spherical` on ingest)

zeroshade:
## Summary

Adds validation-suite coverage for geo behaviors the driver implements but that the baseline geo cases (added in #159) don't exercise. Closes #143.

## What's added

**SELECT — `go/validation/queries/type/select/`**
- `geometry_ewkb_no_srid.txtcase` — GEOMETRY with `geometry_output_format=EWKB` and no SRID set; asserts the `geoarrow.wkb` column carries **no** `crs` metadata (the `analyzeGeoFromBatch` "no consistent SRID" path).
- `geometry_ewkb_mixed_srid.txtcase` — rows with differing SRIDs (2029 / 4326); asserts the column is still detected as geo but emits **no** `crs` metadata (the `mixed` path).

**INGEST — `go/validation/queries/ingest/`**
- `geoarrow_wkb_geometry_autodetect.txtcase` / `geoarrow_wkb_geography_autodetect.txtcase` — same inputs as the existing WKB ingest cases but with **no** `ingest_geo_type` option set, exercising auto-detection (`EPSG:4326`+spherical → GEOGRAPHY; non-4326 → GEOMETRY).
- `geoarrow_wkt_geometry.txtcase` / `geoarrow_wkt_geography.txtcase` — WKT (`geoarrow.wkt`) ingest input, which previously had no coverage.

**Unit test — `go/statement_test.go`**
- Adds empty / `{}` (no-CRS) cases to the existing `TestResolveGeoType` (the auto-detect matrix, including the non-4326+spherical error, was already covered).

## Verification

- `go test ./ -run TestResolveGeoType` and `go vet ./...` pass locally.
- All six new `.txtcase` files are parsed and collected by the validation harness (`pytest --collect-only`) with no errors.
- The end-to-end `.txtcase` assertions run against a live Snowflake via `pixi run validate`, which needs credentials not available in my environment. Their expected/golden values were **derived from the existing verified geo cases** — identical geometries, with only CRS-metadata presence (SELECT) and the WKB→WKT input encoding (ingest) differing — so the credentialed CI run is the final gate. Please confirm `pixi run validate` is green before merge.


go/validation/queries/ingest/geoarrow_wkb_geometry_autodetect.txtcase
```diff
@@ -0,0 +1,86 @@
...
+        },
+        {
+            "name": "value",
+            "format": "u",
+            "flags": ["nullable"]
+        }
+    ]
+}
+
+// part: expected
```
lidavidm, line 76: This needs to be fixed

zeroshade: These geo ingest golden values were derived from the GEOMETRY cases, so the GEOGRAPHY outputs used the full-precision float form (`1.000000000000000e+00`) instead of the compact GeoJSON Snowflake returns for GEOGRAPHY (`1`, cf. `type/select/geography.txtcase`).

The case that actually failed the validation suite was `geoarrow_wkt_geography`, now corrected to the compact form in 97682b7. The values in this file (`geoarrow_wkb_geometry_autodetect`) are GEOMETRY output, match the verified `geoarrow_wkb_geometry` case, and pass the suite, so I left them unchanged — let me know if you were flagging something else here.

lidavidm (approved):

PR #165, opened 2026-06-30
fix(go): handle UUID data type in metadata paths

### LINKED ISSUE #164: fix(go): UUID data type not handled in metadata paths (GetTableSchema / GetObjects)
## What happened

Columns of Snowflake's [`UUID` data type](https://docs.snowflake.com/en/sql-reference/data-types-uuid) cannot be discovered through the driver's metadata paths, even though they query fine.

- `GetTableSchema` (which runs `DESC TABLE` → `descToField`) returns a hard error:
  ```
  Not Implemented: Snowflake Data Type UUID not implemented
  ```
- `GetObjects` → `toArrowField` matches no `case` for `UUID`, so the column's `arrow.Field.Type` is left `nil` and no `XdbcDataType`/`XdbcSqlDataType` is populated.

The data path itself is fine: `SELECT * FROM <table>` returns the `UUID` column as Arrow `utf8` (Snowflake delivers UUID values as text over the wire).

This surfaces in the Power BI Snowflake connector (built on this ADBC driver) as a Navigator error when previewing/loading a table that has a `UUID` column:

```
Expression.Error: Unable to understand the type for column
Details:
    <UUID column name>
```

## Reproduction

```sql
create table tester (uuid_id UUID, varchar_id VARCHAR);
insert into tester (uuid_id, varchar_id) select uuid_string(), uuid_string();
```

Then, against `TESTER`:

| Path | Result |
| --- | --- |
| `GetTableSchema` | ❌ `Not Implemented: Snowflake Data Type UUID not implemented` |
| `GetObjects` (column depth) | ⚠️ field type left `nil` |
| `SELECT *` result schema | ✅ `utf8` |

`SHOW COLUMNS` reports the type as `{"type":"UUID","nullable":true}`; `DESC TABLE` reports it as `UUID`.

## Expected

The metadata paths should agree with the data path and report a `UUID` column as Arrow `utf8`, consistent with how `VARIANT`/`OBJECT`/`ARRAY` are surfaced.

## Environment

- `go/` driver, `github.com/snowflakedb/gosnowflake/v2 v2.1.0`

prmoore77:
## Problem

Snowflake's [`UUID` data type](https://docs.snowflake.com/en/sql-reference/data-types-uuid) queries fine but was unhandled in the driver's two schema-discovery paths:

- **`GetTableSchema`** (`DESC TABLE` → `descToField`): a parenless `UUID` type fell into the `default` and returned a hard error — `Not Implemented: Snowflake Data Type UUID not implemented`.
- **`GetObjects`** (`SHOW COLUMNS` → `toArrowField`): no `case` matched `UUID`, so `arrow.Field.Type` was left `nil`.

This surfaced downstream (e.g. the Power BI Snowflake connector, Metabase) as failures like `Expression.Error: Unable to understand the type for column`.

## Fix

UUID columns are surfaced as `arrow.uuid`.

## How UUID result columns are detected

Snowflake does **not** report a dedicated `uuid` type in query-result metadata. Verified against a live account: both the rowtype JSON and the Arrow field metadata report UUID result columns as `text`/`TEXT` (consistent with the docs: "Snowflake drivers treat UUID values as text strings"). Only `DESC TABLE` / `SHOW COLUMNS` name the type — so those paths map `UUID` directly.

On the result path, UUID columns are identified by their otherwise-impossible shape: **`text` with length 0** (VARCHAR's minimum length is 1; SHOW commands, empty strings, `NULL::VARCHAR`, concatenations, etc. all report length ≥ 1, while table columns, `NULL::UUID`, and UUID-typed expressions all report 0). The one other length-0 producer — an untyped `NULL` literal (`SELECT NULL`) — is ruled out on the Arrow data path via its explicit `byteLength: "0"` field metadata (UUID columns carry `"16"` or omit the key). A literal `uuid` rowtype is also matched, to future-proof against the server reporting the logical type directly.

## Changes

- `descToField` (`GetTableSchema`) and `toArrowField` (`GetObjects`) report `arrow.uuid`
- the Arrow data path parses the utf8 values into an `extensions.UUIDArray` via a column transformer; the JSON fallback path types the field so the extension builder parses the text
- **Bulk ingest**: `arrow.uuid` columns are created as native `UUID` columns (previously `BINARY(16)`). On the plain COPY path no transform is needed — Snowflake's Parquet reader converts the Parquet UUID logical type (emitted for arrow.uuid's `fixed_size_binary[16]` storage) natively, verified value-for-value against a live account. On the geo COPY-transform path (where every column is read through a `$1:"col"` VARIANT reference), the 16 raw bytes are re-encoded to canonical UUID text inline.
- **Bind**: `arrow.uuid` parameters (both registered `extensions.UUIDArray` values and C-Data-Interface arrays that arrive as `fixed_size_binary[16]` storage plus `ARROW:extension:name` field metadata) bind as canonical UUID text, which Snowflake converts implicitly.
- Extension detection follows the existing geoarrow pattern: both `arrow.EXTENSION` types and `ARROW:extension:name` field metadata are recognized.

Closes #164

### DIFF
--- a/go/driver_test.go
+++ b/go/driver_test.go
@@ -1888,6 +1888,48 @@ func (suite *SnowflakeTests) TestGeometryAsText() {
 	suite.Contains(geogCol.Value(0), "Point")
 }
 
+func (suite *SnowflakeTests) TestUUIDType() {
+	// Snowflake's UUID data type is delivered as text over the wire, so it must
+	// surface as Arrow utf8 in every path. Before this was handled, the metadata
+	// paths failed ("Snowflake Data Type UUID not implemented" / nil field type),
+	// which surfaced downstream (e.g. Power BI) as "Unable to understand the type
+	// for column".
+	suite.Require().NoError(suite.stmt.SetSqlQuery(suite.ctx, `CREATE OR REPLACE TABLE UUID_TYPE_TEST (
+		UUID_ID UUID,
+		VARCHAR_ID VARCHAR
+	)`))
+	_, err := suite.stmt.ExecuteUpdate(suite.ctx)
+	suite.Require().NoError(err)
+
+	suite.Require().NoError(suite.stmt.SetSqlQuery(suite.ctx,
+		"INSERT INTO UUID_TYPE_TEST SELECT uuid_string(), uuid_string()"))
+	_, err = suite.stmt.ExecuteUpdate(suite.ctx)
+	suite.Require().NoError(err)
+
+	// Metadata path: GetTableSchema (DESC TABLE -> descToField).
+	cat, sch := suite.Quirks.catalogName, suite.Quirks.schemaName
+	tableSchema, err := suite.cnxn.GetTableSchema(suite.ctx, &cat, &sch, "UUID_TYPE_TEST")
+	suite.Require().NoError(err)
+	suite.Truef(arrow.TypeEqual(arrow.BinaryTypes.String, tableSchema.Field(0).Type),
+		"GetTableSchema UUID_ID: expected utf8, got %s", tableSchema.Field(0).Type)
+
+	// Data path: result schema must match, and values must round-trip.
+	suite.Require().NoError(suite.stmt.SetSqlQuery(suite.ctx, "SELECT * FROM UUID_TYPE_TEST"))
+	rdr, _, err := suite.stmt.ExecuteQuery(suite.ctx)
+	suite.Require().NoError(err)
+	defer rdr.Release()
+
+	uuidField := rdr.Schema().Field(0)
+	suite.Truef(arrow.TypeEqual(arrow.BinaryTypes.String, uuidField.Type),
+		"SELECT UUID_ID: expected utf8, got %s", uuidField.Type)
+
+	suite.True(rdr.Next())
+	rec := rdr.RecordBatch()
+	uuidCol := rec.Column(0).(*array.String)
+	_, parseErr := uuid.Parse(uuidCol.Value(0))
+	suite.NoErrorf(parseErr, "UUID_ID value %q should parse as a UUID", uuidCol.Value(0))
+}
+
 func (suite *SnowflakeTests) TestTimestampPrecisionJson() {
 	opts := suite.Quirks.DatabaseOptions()
 	opts[driver.OptionMaxTimestampPrecision] = "microseconds"
--- a/go/driver_test.go
+++ b/go/driver_test.go
@@ -53,6 +53,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow"
 	"github.com/apache/arrow-go/v18/arrow/array"
 	"github.com/apache/arrow-go/v18/arrow/decimal128"
+	"github.com/apache/arrow-go/v18/arrow/extensions"
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/google/uuid"
 	"github.com/snowflakedb/gosnowflake/v2"
@@ -1889,11 +1890,14 @@ func (suite *SnowflakeTests) TestGeometryAsText() {
 }
 
 func (suite *SnowflakeTests) TestUUIDType() {
-	// Snowflake's UUID data type is delivered as text over the wire, so it must
-	// surface as Arrow utf8 in every path. Before this was handled, the metadata
-	// paths failed ("Snowflake Data Type UUID not implemented" / nil field type),
-	// which surfaced downstream (e.g. Power BI) as "Unable to understand the type
-	// for column".
+	// Snowflake delivers UUID values as text over the wire; the driver decodes
+	// them into the arrow.uuid extension type (fixed_size_binary[16] storage)
+	// in both the metadata and data paths. Before this was handled, the
+	// metadata paths failed ("Snowflake Data Type UUID not implemented" / nil
+	// field type), which surfaced downstream (e.g. Power BI) as "Unable to
+	// understand the type for column".
+	uuidType := extensions.NewUUIDType()
+
 	suite.Require().NoError(suite.stmt.SetSqlQuery(suite.ctx, `CREATE OR REPLACE TABLE UUID_TYPE_TEST (
 		UUID_ID UUID,
 		VARCHAR_ID VARCHAR
@@ -1910,8 +1914,8 @@ func (suite *SnowflakeTests) TestUUIDType() {
 	cat, sch := suite.Quirks.catalogName, suite.Quirks.schemaName
 	tableSchema, err := suite.cnxn.GetTableSchema(suite.ctx, &cat, &sch, "UUID_TYPE_TEST")
 	suite.Require().NoError(err)
-	suite.Truef(arrow.TypeEqual(arrow.BinaryTypes.String, tableSchema.Field(0).Type),
-		"GetTableSchema UUID_ID: expected utf8, got %s", tableSchema.Field(0).Type)
+	suite.Truef(arrow.TypeEqual(uuidType, tableSchema.Field(0).Type),
+		"GetTableSchema UUID_ID: expected arrow.uuid, got %s", tableSchema.Field(0).Type)
 
 	// Data path: result schema must match, and values must round-trip.
 	suite.Require().NoError(suite.stmt.SetSqlQuery(suite.ctx, "SELECT * FROM UUID_TYPE_TEST"))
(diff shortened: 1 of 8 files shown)


CurtHagenlocher:
> The Power BI M-layer type mapper handles utf8 but not fixed_size_binary/extension types

That could be addressed.

I feel like the right thing to do for the driver should be independent of a particular consumption scenario. And driver changes should anyway be (ideally) backwards-compatible, e.g. by opting into new behavior with a connection property.

prmoore77:
This is a screenshot of the Power BI connector displaying the error: 
image

zeroshade:
It's frustrating that Snowflake stores the UUID values as fixed size binary values, but still uses the arrow String type for how it returns it :frowning:

CurtHagenlocher:
> This is a screenshot of the Power BI connector displaying the error:

We, um, don't use the ADBC functions to get the table schema. We run `SHOW COLUMNS IN TABLE` and parse the results ourselves, so this is totally a bug in our connector and not the driver.

That's not to say that the driver couldn't be improved in this fashion... .

prmoore77:
> > This is a screenshot of the Power BI connector displaying the error:
> 
> We, um, don't use the ADBC functions to get the table schema. We run `SHOW COLUMNS IN TABLE` and parse the results ourselves, so this is totally a bug in our connector and not the driver.
> 
> That's not to say that the driver couldn't be improved in this fashion... .

Hi, @CurtHagenlocher - Is the connector open-source?  Should I file an issue in that repo (if it is)?

FWIW - We also observed a similar issue in Metabase - and they had to make a fix as well: https://github.com/metabase/metabase/pull/72034

CurtHagenlocher:
> Hi, @CurtHagenlocher - Is the connector open-source? Should I file an issue in that repo (if it is)?

Alas, it is not. I've filed 2175707 in our internal tracking system.

go/driver_test.go
lidavidm, line 1: This should be added as a validation suite case

lidavidm (commented):
Is this another case where Snowflake doesn't tell us the "real" logical type and so we're unable to attach the proper extension type?

zeroshade:
> Is this another case where Snowflake doesn't tell us the "real" logical type and so we're unable to attach the proper extension type?

It looks like snowflake *does* tell us the "real" logical type in the Column Info and descriptions (but not in the arrow schema) as "UUID", but it still delivers the UUID explicitly as a string and not as a fixed_size_binary[16].

lidavidm:
> > Is this another case where Snowflake doesn't tell us the "real" logical type and so we're unable to attach the proper extension type?
> 
> It looks like snowflake _does_ tell us the "real" logical type in the Column Info and descriptions (but not in the arrow schema) as "UUID", but it still delivers the UUID explicitly as a string and not as a fixed_size_binary[16].

In that case, though, presumably it's not too expensive to cast the data and "fix" it?

zeroshade:
> In that case, though, presumably it's not too expensive to cast the data and "fix" it?

That would be my preference also, but it looks like PowerBI rejects that? Do I have that right @prmoore77 @CurtHagenlocher ? 

Is it possible to change this so that we cast the uuid values to a proper arrow.uuid extension type using fixed size binary[16]?

prmoore77:
> > In that case, though, presumably it's not too expensive to cast the data and "fix" it?
> 
> That would be my preference also, but it looks like PowerBI rejects that? Do I have that right @prmoore77 @CurtHagenlocher ?
> 
> Is it possible to change this so that we cast the uuid values to a proper arrow.uuid extension type using fixed size binary[16]?

Hey @zeroshade - that is fine from my side.  I can close this PR if you would like to make the fix you would recommend - or I could adjust it, but you would know better than I do about how it should be done...

zeroshade:
I'm fine with you adjusting this PR to use the extension type and cast the string we get from Snowflake accordingly

prmoore77:
Hi, @zeroshade @lidavidm - Done — the PR now surfaces UUID columns as the `arrow.uuid` extension type (`fixed_size_binary[16]`) and casts the text values from the wire into it, and the test moved into the validation suite as `queries/type/select/uuid.txtcase`. One important nuance I hit while implementing it:

> It looks like snowflake *does* tell us the "real" logical type in the Column Info and descriptions (but not in the arrow schema) as "UUID"

That turns out to be true only for `DESC TABLE` / `SHOW COLUMNS` (so `GetTableSchema`/`GetObjects` map `UUID` directly). On the **query-result path** there is no `uuid` type anywhere: I verified against a live account that both the rowtype JSON and the Arrow field metadata report UUID result columns as `text`/`TEXT` — consistent with the [UUID docs](https://docs.snowflake.com/en/sql-reference/data-types-uuid): "Snowflake drivers treat UUID values as text strings."

What UUID result columns *do* have is an otherwise-impossible shape: **`text` with length 0**. VARCHAR's minimum length is 1, and I probed a pile of text producers (SHOW commands, empty strings, `NULL::VARCHAR`, `TO_CHAR`, concatenations, `GET_DDL`, …) — all report length ≥ 1, while UUID table columns, `NULL::UUID`, and UUID-typed expressions all report 0. So the data path detects UUID as `text && length == 0` (plus a literal `uuid` rowtype match, to future-proof against the server ever reporting the logical type directly).

The one collision is an untyped `NULL` literal (`SELECT NULL`), which also comes back as `text` length 0. Those are excluded on the Arrow data path via the server's field metadata: untyped NULL carries an explicit `byteLength: "0"`, while UUID columns carry their 16-byte storage width (`"16"`) or omit the key. There's a regression check in `TestUUIDType` asserting `SELECT NULL` stays `utf8`.

I'll grant this is a fingerprint rather than a documented contract — but it's the only signal the server gives us today (similar in spirit to the EWKB sniffing the driver already does for geo columns), and if gosnowflake/the server ever start reporting `uuid` directly, the explicit type match takes over.

The round trip is also covered now: bulk ingest creates native `UUID` columns for `arrow.uuid` data (previously `BINARY(16)`) — no COPY transform needed on the plain path, since Snowflake's Parquet reader converts the Parquet UUID logical type natively — and `arrow.uuid` bind parameters (previously rejected) bind as canonical UUID text. Both come with validation-suite cases (`ingest/uuid`, `type/bind/uuid`) in addition to `type/select/uuid`.

Verified live: the UUID validation cases pass 8/8, the full validation query suite passes 212/212, and the ingest and integration suites show no new failures beyond ones that reproduce identically on `main`.

prmoore77:
Hi @zeroshade and @lidavidm - there is a failing test above, but I think it is due to credentials.

zeroshade:
yea, since this PR is on a fork it can't access the secrets to run it which is why it's failing. I'll run it myself manually to confirm and post back

zeroshade (approved):
running locally, the validation on this branch passed successfully. I'll wait for @lidavidm to comment before merging, but this looks good to me

prmoore77:
Hi @zeroshade and @lidavidm - is there any chance this fix could get merged in some form?  Having UUID's available in Snowflake via ADBC would be beneficial...  Thanks for your help.

prmoore77:
> > Hi, @CurtHagenlocher - Is the connector open-source? Should I file an issue in that repo (if it is)?
> 
> Alas, it is not. I've filed 2175707 in our internal tracking system.

Hi @CurtHagenlocher - is that issue being worked at all?  (just curious - asking for a friend 😊)

go/validation/queries/type/select/uuid.txtcase
```diff
@@ -0,0 +1,67 @@
...
+;
+
+// part: query
+
+SELECT RES FROM TEST_UUID ORDER BY IDX ASC
+
+// part: expected_schema
+// Snowflake delivers UUID values as utf8 text on the wire; the driver decodes
+// them into the arrow.uuid extension type (fixed_size_binary[16] storage) and
+// reports the same type from the metadata paths.
```
lidavidm:
```suggestion
```

Can we clean up some of the AI comments?

go/driver_test.go
lidavidm: ...and now that the cases are added, is this still necessary?

go/record_reader.go
```diff
@@ -165,6 +166,53 @@ func stripEWKBSRIDColumn(_ context.Context, a arrow.Array) (arrow.Array, error)
 	return bldr.NewArray(), nil
 }
 
+// isUUIDResultColumn reports whether a result column carries Snowflake UUID
+// values. Snowflake result metadata does not surface a dedicated "uuid" type
+// today: UUID columns are reported as "text" but with length 0, which no real
+// text column can have (VARCHAR's minimum length is 1, and even SHOW commands
+// report the maximum length for their text columns). The explicit "uuid"
+// match future-proofs against the server reporting the logical type directly.
```
lidavidm: If this is never actually used, I don't think it makes sense to check for it. There's no guarantee the implementation would look like this anyways.

go/record_reader.go
```diff
@@ -600,6 +657,13 @@ func rowTypesToArrowSchema(_ context.Context, ld gosnowflake.ArrowStreamLoader,
 				[]string{srcMeta.Type},
 			),
 		}
+		if isUUIDResultColumn(srcMeta.Type, srcMeta.Length) {
+			// jsonDataToArrow appends the JSON UUID text through the
+			// extension builder's AppendValueFromString, which parses it
+			// into the fixed_size_binary[16] storage.
```
lidavidm:
```suggestion
```

go/record_reader.go
```diff
@@ -298,6 +346,15 @@ func getTransformer(sc *arrow.Schema, ld gosnowflake.ArrowStreamLoader, useHighP
 			continue
 		}
 
+		// Snowflake delivers UUID values as utf8 text on the wire; decode
+		// them into the canonical arrow.uuid extension type.
```
lidavidm:
```suggestion
```

lidavidm:
(Also, if we use the [workflow](https://github.com/adbc-drivers/snowflake/actions/workflows/go_test_pr.yaml), then the test results will be posted to the PR.)

CurtHagenlocher:
> Hi @CurtHagenlocher - is that issue being worked at all? (just curious - asking for a friend 😊)

I'm going to try to get to it this week.

go/record_reader.go
prmoore77: Removed the speculative uuid type-name match; the function now only checks the shape Snowflake actually reports today

prmoore77:
@lidavidm - for `driver_test.go`: Removed both integration tests as redundant with the validation cases; kept only a small TestUntypedNullLiteral regression test for the SELECT NULL edge case, which the validation  
  suite can't express.
  
  (I wasn't sure how to reply in-thread for this)

lidavidm (approved):

prmoore77:
🚀

PR #167, opened 2026-07-04
fix(go): geo type auto-detect should not require edges:spherical for GEOGRAPHY

jatorre:
`resolveGeoType`'s own doc comment says *"missing CRS, EPSG:4326, or unparsable CRS stays GEOGRAPHY"*, and `docs/snowflake.md`'s option table described detection differently again — but the implementation required `edges == "spherical"` in addition to SRID 4326 before choosing GEOGRAPHY.

Most GeoArrow producers emit a `crs` but **no `edges` field** (DuckDB's spatial extension and GeoPandas among them), so in practice every WGS84 column ingested through auto-detect silently landed as GEOMETRY. We hit this in CARTO's integration suite when moving from the pre-merge #117 branch to `go/v1.11.0`: same import, `GEOGRAPHY` became `GEOMETRY`.

This aligns the code with the documented contract:

- SRID 4326 **or no CRS** → GEOGRAPHY
- any other SRID → GEOMETRY (SRID survives the round trip via ST_SETSRID)
- `edges:spherical` with a non-4326 SRID → still an error (unchanged)
- explicit `adbc.snowflake.statement.ingest_geo_type` → unchanged, pins all columns

Updated `TestResolveGeoType` and the option docs to match. The two auto-detect validation txtcases are unaffected (the GEOGRAPHY one uses 4326+spherical, the GEOMETRY one uses EPSG:2029).

https://claude.ai/code/session_01Y3xqZXWrx9qXdPNXBzNxa2

### DIFF
--- a/go/docs/snowflake.md
+++ b/go/docs/snowflake.md
@@ -366,7 +366,7 @@ Examples:
 `adbc.snowflake.statement.ingest_geo_type`
 : **Values:** `geography`, `geometry`, or empty string. **Default:** empty string
 
-  Which Snowflake data type to use when ingesting columns of GeoArrow extension types (`geoarrow.wkb`, `geoarrow.wkt`). If empty, then it will be detected: if the SRID is 4326 and `edges:spherical` is set, then it will be ingested as GEOGRAPHY, else GEOMETRY. If explicitly set, the driver will use the specified type.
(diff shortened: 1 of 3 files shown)


lidavidm:
CC @paleolimbot but my understanding is that this was semantically required

PR #170, opened 2026-07-13
fix(go): scope SHOW COLUMNS to the requested object in GetObjects

### LINKED ISSUE #169: GetObjects at Columns depth runs SHOW COLUMNS IN ACCOUNT (80-150s) for underscore catalog names
## Summary

`GetObjects` at `ObjectDepthColumns` / `ObjectDepthAll` can take 80–150 seconds for a single-table request when the catalog (database) name contains an underscore.

Reported on #105, but the root cause is unrelated to that PR.

## Symptoms

Calling `GetObjects(depth=Columns, catalog="DB_NAME", dbSchema="TEST_SCHEMA", tableName="LINEITEM")` takes ~90–150s, regardless of how many columns the target table has (reproduced with a 1-column table). The equivalent call was ~10s with the older `apache/arrow-adbc/go/adbc` driver.

`QUERY_HISTORY` from a ~92s call shows the time is almost entirely one query:

| SHOW query (`/* ADBC:getObjects */`) | Elapsed |
|---|---|
| `SHOW TERSE DATABASES LIKE 'DB_NAME' IN ACCOUNT` | 179 ms |
| `SHOW UNIQUE KEYS ... IN ACCOUNT` | 406 ms |
| `SHOW IMPORTED KEYS ... IN ACCOUNT` | 267 ms |
| **`SHOW COLUMNS /* ADBC:getObjects */ IN ACCOUNT`** | **84,866 ms** |
| `SHOW TERSE SCHEMAS LIKE 'TEST_SCHEMA' IN ACCOUNT` | 333 ms |

## Root cause

`GetObjects` builds the scope suffix for `SHOW COLUMNS` from `isWildcardStr`, which treats both `_` and `%` as wildcards:

```go
func isWildcardStr(ident string) bool {
    return strings.ContainsAny(ident, "_%")
}
```

For an ordinary database name like `DB_NAME`, the catalog is considered a wildcard, so the scope collapses to `IN ACCOUNT`. Unlike the other SHOW commands, `SHOW COLUMNS` has no table-level `LIKE` filter, so `SHOW COLUMNS ... IN ACCOUNT` scans **every column of every table in every database in the account**. On a large account that is the 85s above; the target table's column count is irrelevant.

The `SHOW COLUMNS` call was introduced in #152 (`fix(go): set xdbc_column_size for BINARY columns in GetObjects`) to read `byteLength` for BINARY column sizes. The older `apache/arrow-adbc` driver read column metadata solely from `information_schema.columns` (scoped by catalog in SQL) and never issued a `SHOW COLUMNS`, which is why the regression appears when moving to this driver.

Scope timing measured on a small test account (worse on large accounts):

| `SHOW COLUMNS` scope | Elapsed |
|---|---|
| `IN ACCOUNT` | 18.4 s |
| `IN DATABASE` | 4.8 s |
| `IN SCHEMA` | 0.52 s |
| `IN TABLE` | 0.39 s |

## Proposed fix

Scope the `SHOW COLUMNS` query to the narrowest concrete object (`IN TABLE` / `IN SCHEMA` / `IN DATABASE`), treating `_` as a literal and only `%` / `.*` / nil as broadening. This is safe because `SHOW COLUMNS` is only a best-effort source for BINARY `xdbc_column_size`; the authoritative column list still comes from `information_schema.columns` filtered by `ILIKE`, so a narrower scope never drops columns or changes non-BINARY metadata.

Tradeoff: a BINARY column reached only through a genuine `_`-as-wildcard match may report a NULL `xdbc_column_size`, matching the behavior before #152.

## Environment

- Driver: `github.com/adbc-drivers/snowflake/go` with `github.com/snowflakedb/gosnowflake/v2`
- Previously fast: `github.com/apache/arrow-adbc/go/adbc` (v1.11.0) with `github.com/snowflakedb/gosnowflake` (v1.19.1)

zeroshade:
## Summary

Fixes #169.

`GetObjects` at `ObjectDepthColumns` / `ObjectDepthAll` dispatched a `SHOW COLUMNS` command whose scope resolved to `IN ACCOUNT` whenever the catalog name contained an underscore, because `isWildcardStr` treats `_` and `%` as wildcards. Since `SHOW COLUMNS` has no table-level `LIKE` filter, `SHOW COLUMNS ... IN ACCOUNT` scans every column in the entire account — 80–150s on large accounts — even for a single-table request like `DB_NAME.TEST_SCHEMA.LINEITEM`.

## Changes

- Add `scopeIdentifier` / `showColumnsScope` and scope the `SHOW COLUMNS` query to the narrowest concrete object (`IN TABLE` / `IN SCHEMA` / `IN DATABASE`), treating `_` as a literal and only `%` / `.*` / nil as broadening.
- Limit the `2003` (object does not exist / not authorized) empty-result fallback to only the optional `SHOW COLUMNS` enrichment query, via a new variadic `alsoEmptyOn` parameter on `getQueryID`. Required object/constraint discovery keeps its prior `2043`-only behavior.
- Add a unit test for scope selection.

## Why this is safe

`SHOW COLUMNS` is only a best-effort source for BINARY `xdbc_column_size` in `get_objects_all.sql` (a `LEFT JOIN` for `byte_length`). The authoritative column list comes from `information_schema.columns` filtered by `ILIKE`, so a narrower `SHOW COLUMNS` scope never drops columns or alters non-BINARY metadata.

**Tradeoff:** a BINARY column reached only through a genuine `_`-as-wildcard match (or a case-insensitive pattern differing from the stored identifier) may report a NULL `xdbc_column_size`, matching the behavior before `SHOW COLUMNS` enrichment was added in #152. This is documented on `scopeIdentifier` / `showColumnsScope`.

## Scope timing (small test account; worse on large accounts)

| `SHOW COLUMNS` scope | Elapsed |
|---|---|
| `IN ACCOUNT` | 18.4 s |
| `IN DATABASE` | 4.8 s |
| `IN SCHEMA` | 0.52 s |
| `IN TABLE` | 0.39 s |

## Testing

- `go build`, `go vet`, `gofmt`: clean.
- `TestShowColumnsScope`: passes (concrete underscore names → `IN TABLE`; `%` / `.*` / nil → broader; quote escaping).
- Live check against a real account: `GetObjects(Columns, "SNOWFLAKE_SAMPLE_DATA", "TPCH_SF1", "LINEITEM")` returns correctly and drops from the ~18s `IN ACCOUNT` floor to a table-scoped query; a non-existent table returns an empty result rather than erroring.

### DIFF
--- a/go/connection.go
+++ b/go/connection.go
@@ -125,8 +125,10 @@ func getQueryID(ctx context.Context, query string, driverConn driver.QueryerCont
 	rows, err := driverConn.QueryContext(ctx, query, nil)
 	if err != nil {
 		var sfErr *gosnowflake.SnowflakeError
-		// No results. Generate a dummy result set instead
-		if emptyQuery != "" && errors.As(err, &sfErr) && sfErr.Number == 2043 {
+		// 2043: the SHOW command matched nothing. 2003: the scoped object does
+		// not exist or is not authorized. In both cases substitute an empty
+		// result set so RESULT_SCAN has a valid (empty) source to read from.
+		if emptyQuery != "" && errors.As(err, &sfErr) && (sfErr.Number == 2043 || sfErr.Number == 2003) {
 			return getQueryID(ctx, emptyQuery, driverConn, "")
 		}
 		return "", err
@@ -191,6 +193,43 @@ func goGetQueryID(ctx context.Context, conn driver.QueryerContext, grp *errgroup
 
 func isWildcardStr(ident string) bool {
 	return strings.ContainsAny(ident, "_%")
+}
+
+// scopeIdentifier reports whether ident can be used as a concrete object name to
+// narrow a SHOW command's scope (IN DATABASE/SCHEMA/TABLE). Only the
+// multi-character wildcard '%' and the match-all sentinels force a broader
+// scope; a '_' is treated as a literal character. Treating '_' as a wildcard
+// here would scope SHOW COLUMNS to IN ACCOUNT for ordinary names like
+// "DB_NAME", scanning every column in the entire account; the downstream ILIKE
+// filtering in the query template still applies the pattern semantics.
+func scopeIdentifier(ident *string) (string, bool) {
+	if ident == nil {
+		return "", false
+	}
+	s := *ident
+	if s == "" || s == "%" || s == ".*" || strings.Contains(s, "%") {
+		return "", false
+	}
+	return s, true
+}
+
+// showColumnsScope returns the narrowest "IN ..." suffix for the SHOW COLUMNS
+// query given the requested catalog/schema/table, avoiding an account-wide
+// column scan whenever concrete names are provided.
+func showColumnsScope(catalog, dbSchema, tableName *string) string {
+	cat, ok := scopeIdentifier(catalog)
+	if !ok {
+		return " IN ACCOUNT"
+	}
+	sch, ok := scopeIdentifier(dbSchema)
+	if !ok {
+		return " IN DATABASE " + quoteIdentifier(cat)
+	}
+	tbl, ok := scopeIdentifier(tableName)
+	if !ok {
+		return " IN SCHEMA " + quoteIdentifier(cat) + "." + quoteIdentifier(sch)
+	}
+	return " IN TABLE " + quoteIdentifier(cat) + "." + quoteIdentifier(sch) + "." + quoteIdentifier(tbl)
 }
 
 func (c *connectionImpl) GetObjects(ctx context.Context, depth adbc.ObjectDepth, catalog, dbSchema, tableName, columnName *string, tableType []string) (reader array.RecordReader, err error) {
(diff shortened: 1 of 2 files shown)


zeroshade:
@frbvianna can you take a look and test this out with your config?

lidavidm (commented):
We should probably specify the option(s) to disable and/or escape wildcards for these pattern arguments (upstream)

zeroshade:
I agree, but I don't think that should block this change, should it?

lidavidm:
Didn't say it was blocking :) But ideally @frbvianna can confirm that this actually fixes the problem.

zeroshade:
I was able to reproduce the issue reported and prove this fixed it for myself, but I agree that it would be better for @frbvianna to confirm

frbvianna:
Hey @zeroshade. Thanks for the fix. I gave it a test, and now it's taking a reasonable ~12 seconds to get objects at column depth.

PR #171, opened 2026-07-14
feat(csharp): native C# Snowflake ADBC driver (REST-based)

ndglover:
Adds a from-scratch, fully-managed C# implementation of the Snowflake ADBC driver (net8.0), independent of the Go/cgo Interop driver. It speaks Snowflake's internal REST protocol directly and returns Arrow natively.

As discussed in our slack [chat](https://columnar-community.slack.com/archives/C09P47Q69T6/p1782164666560409?thread_ts=1781880663.761709&cid=C09P47Q69T6) here is a first cut of the driver.

Tagging @lidavidm @ianmcook @davidhcoe @CurtHagenlocher


.gitattributes
lidavidm, line 1: I thought certain IDE files used CRLF?

CurtHagenlocher:
Thanks @lidavidm for leaving a comment which reminded me of this PR. I'll set some time aside tomorrow evening to look at it. There are a bunch of linter issues around whitespace and at least one file missing a copyright notice. I also wonder whether the copyright year should be 2026 ;). And this is probably just a "me" thing, but I find I keep getting confused by the use of "Native" to refer to the managed implementation (being "native C#").

csharp/src/Native/SnowflakeStatement.cs
```diff
@@ -0,0 +1,342 @@
...
+    /// <param name="batch">The RecordBatch containing parameter values.</param>
+    /// <param name="schema">The schema of the RecordBatch.</param>
+    public override void Bind(RecordBatch batch, Schema schema)
+    {
+        ThrowIfDisposed();
+
+        ArgumentNullException.ThrowIfNull(batch);
+        ArgumentNullException.ThrowIfNull(schema);
+
+        _boundParameters = batch;
```
lidavidm, line 98: If _boundParameters is already set, should we dispose it?

csharp/src/Native/Configuration/ConnectionStringParser.cs
```diff
@@ -0,0 +1,314 @@
...
+            Role = GetOptionalParameter(parameters, "adbc.snowflake.sql.role"),
+            QueryTag = GetOptionalParameter(parameters, SnowflakeStatement.QueryTagOption),
+            Authentication = ParseAuthenticationConfig(parameters)
+        };
+
+        if (parameters.TryGetValue("adbc.snowflake.sql.client_option.request_timeout", out string? requestTimeoutStr) &&
+            int.TryParse(requestTimeoutStr, out int requestTimeoutSeconds))
+        {
+            config.QueryTimeout = TimeSpan.FromSeconds(requestTimeoutSeconds);
+        }
```
lidavidm: If the value is present but does not parse, presumably we should error?

.gitattributes
ndglover: I found that by not fixing it to one or the other you end up with certain IDEs automatically modifying the files. This means you need to manually handle this in git by reverting these changes or you end up in ding dong between different conventions people have. I believe fixing to one works better but open to other suggestions

csharp/src/Native/SnowflakeStatement.cs
ndglover: Agreed, I've fixed this now

csharp/src/Native/Configuration/ConnectionStringParser.cs
ndglover: Yes, this is a better pattern, thanks. I've applied a fix

ndglover:
@CurtHagenlocher - I think I've fixed up the linting issues and missing license headers now.
@lidavidm - that's for your comments, these are all fixed as well now

csharp/src/Native/Configuration/AuthenticationConfig.cs
```diff
@@ -0,0 +1,64 @@
+/*
+* Copyright (c) 2025 ADBC Drivers Contributors
```
CurtHagenlocher, line 2: It's closer to 2027 than 2025 ;)

csharp/test/Interop/SnowflakeTestConfiguration.cs
```diff
@@ -102,6 +102,8 @@ internal class SnowflakeTestConfiguration : TestConfiguration
         /// </summary>
         [JsonPropertyName("roleInfo")]
         public RoleInfo? RoleInfo { get; set; }
+
```
CurtHagenlocher, line 105: nit: any reason for the added whitespace?

csharp/AdbcDrivers.Snowflake.sln
```diff
@@ -16,36 +17,124 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AdbcDrivers.Snowflake.Inter
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AdbcDrivers.Snowflake.Interop.Tests", "test\Interop\AdbcDrivers.Snowflake.Interop.Tests.csproj", "{7E63DF1C-ED91-B9F3-064F-D6779FE1AC4F}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{827E0CD3-B72D-47B6-A68D-7590B98EB39B}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Native", "Native", "{986E768A-9E42-6229-8E82-349DB5D13BDD}"
```
CurtHagenlocher, line 22: I find the use of "native" to mean "natively C#" a bit confusing. In the context of .NET it's usually used to mean the opposite of "managed" -- at least in my world; maybe this isn't common?

csharp/src/Native/Services/TypeConversion/SnowflakeDataType.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// Gets the Snowflake type code.
+    /// </summary>
+    public SnowflakeTypeCode TypeCode => ParseTypeCode(TypeName);
+
+    private static SnowflakeTypeCode ParseTypeCode(string typeName)
+    {
+        return typeName.ToUpperInvariant() switch
+        {
+            "FIXED" or "NUMBER" or "DECIMAL" or "NUMERIC" => SnowflakeTypeCode.Number,
+            "INTEGER" or "INT" or "BIGINT" or "SMALLINT" or "TINYINT" or "BYTEINT" => SnowflakeTypeCode.Integer,
```
CurtHagenlocher, line 64: Not really a review comment, but I'm curious -- under what circumstances Snowflake will return one of these type names?

csharp/src/Native/Services/TypeConversion/TypeConverter.cs
```diff
@@ -0,0 +1,213 @@
...
+
+            SnowflakeTypeCode.Time => TimeType.Nanosecond,
+
+            SnowflakeTypeCode.Timestamp or
+            SnowflakeTypeCode.TimestampNtz => new TimestampType(TimeUnit.Nanosecond, timezone: (string?)null),
+
+            SnowflakeTypeCode.TimestampLtz => new TimestampType(TimeUnit.Nanosecond, timezone: "UTC"),
+
+            // The result decoder stores TIMESTAMP_TZ as its UTC instant (a single Arrow column
+            // cannot carry a per-row offset), so the described type matches: Timestamp[ns] "UTC".
```
CurtHagenlocher, line 69: FWIW, in arrow-dotnet 23 there's an Arrow extension type that stores the rough equivalent of a .NET `DateTimeOffset`. It's `TimestampWithOffsetType`.

csharp/src/Native/Configuration/ConnectionStringParser.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        return BuildConfig(merged);
+    }
+
+    private static ConnectionConfig BuildConfig(IReadOnlyDictionary<string, string> parameters)
+    {
+        var config = new ConnectionConfig
+        {
+            Account = GetRequiredParameter(parameters, "adbc.snowflake.sql.account"),
```
CurtHagenlocher, line 84: Consider exposing the option names as public constants for consumers to use and then having the internals consume from the same place.

csharp/src/Native/Services/ConnectionPool/ConnectionPoolEntry.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Gets the semaphore that enforces the maximum pool size.
+    /// </summary>
+    internal SemaphoreSlim CapacitySemaphore { get; } = new(
+        config.PoolConfig.MaxPoolSize,
+        config.PoolConfig.MaxPoolSize);
+
+    /// <summary>
+    /// Lock object for synchronizing access to idle connections.
+    /// </summary>
+    internal readonly object IdleLock = new();
```
CurtHagenlocher, line 53: Is the `IdleLock` needed given that `IdleConnections` is a concurrent data structure?

csharp/src/Native/Services/ConnectionPool/PooledConnection.cs
```diff
@@ -0,0 +1,126 @@
...
+    public DateTimeOffset LastUsedAt => _lastUsedAt;
+
+    public DateTimeOffset LastHeartbeatAt => _lastHeartbeatAt;
+
+    public bool IsDisposed => _disposed;
+
+    // Keyed on the master-token expiry (not the ~1h session), because a session-expired connection is
+    // still usable via reactive renewal until the master lapses — so the pool should only discard it
+    // once it's truly beyond recovery. Evaluated on the pool's clock so all pool timekeeping shares one.
+    public bool IsTokenExpired => _timeProvider.GetUtcNow() >= AuthToken.MasterExpiresAt;
```
CurtHagenlocher, line 86: Does it make sense to have a fudge factor here? If something is going to expire in half a second, what are the odds we'll still be able to make a request without it failing?

csharp/src/Native/Services/Transport/SnowflakeJsonContext.cs
```diff
@@ -0,0 +1,36 @@
...
+using AdbcDrivers.Snowflake.Native.Services.Query;
+
+namespace AdbcDrivers.Snowflake.Native.Services.Transport;
+
+/// <summary>
+/// Source-generated serializer metadata for the query-protocol wire model (a small, closed set of
+/// types), used by <see cref="RestApiClient"/>. Compile-time generation removes the reflection
+/// warm-up and per-call metadata allocations of the default serializer and keeps this path
+/// trimming/AOT-safe. The login path (<c>SnowflakeLoginClient</c>) intentionally stays on the
+/// reflection-based web defaults — it runs once per connection and relies on case-insensitive
```
CurtHagenlocher, line 27: The use of reflection for the login path may preclude the use of AoT compilation for this code.

csharp/test/Native/Configuration/ConnectionStringParserTests.cs
```diff
@@ -0,0 +1,497 @@
...
+{
+    private static Dictionary<string, string> ParseConnectionString(string connectionString)
+    {
+        var parameters = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        var pairs = connectionString.Split(';', StringSplitOptions.RemoveEmptyEntries);
+
+        foreach (var pair in pairs)
+        {
+            var parts = pair.Split('=', 2);
+            if (parts.Length == 2)
```
CurtHagenlocher, line 38: Should this maybe assert that `parts.Length == 2`?

csharp/src/Native/readme.md
```diff
@@ -0,0 +1,264 @@
...
+.NET 8 Release; default warehouse. Wall-clock over a live network varies heavily between runs —
+in this session the *unchanged* Go driver's 1M-row runs ranged 2.2–9.8 s (sd ≈ 2.5 s) — so all
+comparisons use interleaved native/interop runs and multi-run means, never cross-session
+absolutes.
+
+## Known Limitations
+
+- **Live catalog/schema switching**: catalog/schema are honored at `Connect`; `SetOption` after
+  connect supports only `adbc.connection.autocommit`.
+- **Not implemented**: PUT/GET stage file transfer
```
CurtHagenlocher, line 256: Consider calling out the ADBC bulk ingestion API specifically as unsupported

csharp/src/Native/readme.md
```diff
@@ -0,0 +1,264 @@
...
+- Semi-structured types (VARIANT/OBJECT/ARRAY) are returned as JSON strings; GEOGRAPHY/GEOMETRY as GeoJSON strings.
```
CurtHagenlocher, line 257: Two additional limitations:
Timestamps have a tradeoff between precision and range which is narrower than what Snowflake supports
TIMESTAMP_TZ drops/normalizes the offset

csharp/AdbcDrivers.Snowflake.sln
CurtHagenlocher: See for example https://github.com/adbc-drivers/hiveserver2/tree/main/csharp/src and https://github.com/CurtHagenlocher/bigquery/tree/curth_microsoft/aot-clast-packages/csharp/src (which, to be fair, are me) but also https://github.com/lomakinandrei/NativeWPF and https://github.com/3F/Conari and even https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot -- not to belabor the point ;).

csharp/src/Native/SnowflakeStatement.cs
```diff
@@ -0,0 +1,339 @@
...
+    /// Binds parameters to the statement using a RecordBatch.
+    /// </summary>
+    /// <param name="batch">The RecordBatch containing parameter values.</param>
+    /// <param name="schema">The schema of the RecordBatch.</param>
+    public override void Bind(RecordBatch batch, Schema schema)
+    {
+        ThrowIfDisposed();
+
+        ArgumentNullException.ThrowIfNull(batch);
+        ArgumentNullException.ThrowIfNull(schema);
```
CurtHagenlocher, line 92: I don't remember any more why this function takes a `Schema`. Seems like a wart in the API :/.

csharp/src/Native/Services/Authentication/SnowflakeLoginClient.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            var responseContent = await response.Content.ReadFromJsonAsync<LoginResponse>(cancellationToken).ConfigureAwait(false);
+
+            if (responseContent?.Data == null)
+                throw new AdbcException("Invalid response from Snowflake authentication service.");
+
+            if (!responseContent.Success)
+            {
+                var errorMessage = responseContent.Message ?? "Authentication failed.";
+                throw new AdbcException($"Snowflake authentication failed: {errorMessage}");
```
CurtHagenlocher, line 94: In general, it would be nice to try to assign an `AdbcStatusCode` to the exceptions. This is particularly true for authentication-related exceptions, which a consumer has no way to distinguish from other kinds of exceptions except by examining the text.

csharp/src/Native/Services/Query/ChunkedArrowArrayStream.cs
```diff
@@ -0,0 +1,261 @@
...
+            var arrowBytes = Convert.FromBase64String(rowSetBase64);
+            firstStream = new MemoryStream(arrowBytes);
+            firstReader = new Ipc.ArrowStreamReader(firstStream);
+        }
+        else if (chunkList.Count > 0)
+        {
+            var first = chunkList[0];
+            chunkList.RemoveAt(0);
+            firstStream = await apiClient.GetArrowStreamAsync(first.Url, authToken, chunkHeaders, qrmk, cancellationToken).ConfigureAwait(false);
+            firstReader = new Ipc.ArrowStreamReader(firstStream);
```
CurtHagenlocher, line 84: You should be able to pass a `byte[]` directly to the constructor of an `ArrowStreamReader`; it has an overload that takes a `ReadOnlyMemory<byte>`. This will be more efficient than wrapping the memory in a `Stream`.

csharp/src/Native/Services/Query/ChunkedArrowArrayStream.cs
```diff
@@ -0,0 +1,261 @@
...
+            }
+
+            if (!await _channel.Reader.WaitToReadAsync(cancellationToken).ConfigureAwait(false))
+                return null;
+
+            if (!_channel.Reader.TryRead(out var nextChunk))
+                return null;
+
+            _currentStream = nextChunk.Stream;
+            _currentReader = new Ipc.ArrowStreamReader(_currentStream);
```
CurtHagenlocher, line 237: If you know this is going to be a `MemoryStream` (or even if you suspect it could be) then it might be worth extracting the underlying buffer and converting it to a ReadOnlyMemory<byte> of the appropriate length and passing that to the constructor of `ArrowStreamReader`. (Or at least it's worth benchmarking -- I don't remember exactly what the deserialization tries to do and it's possible that using a `Stream` as a source gives more but smaller allocations in a way that actually ends up better off by avoiding the LOH and/or freeing more memory sooner.)

csharp/src/Native/Services/Query/SnowflakeResultArrowStream.cs
```diff
@@ -0,0 +1,421 @@
...
+                    transforms.Add(new ColumnTransform(i, source => ConvertTime(source, target, columnScale)));
+                    outField = new Field(field.Name, target, field.IsNullable, field.Metadata);
+                    break;
+                }
+
+                case TimestampNtzLogicalType:
+                case TimestampLtzLogicalType:
+                case TimestampTzLogicalType:
+                {
+                    // TZ carries a per-row offset field (dropped — we store the UTC instant); NTZ
```
CurtHagenlocher, line 160: See other comment about the possibility of using the extension type.

csharp/src/Native/Services/Query/SnowflakeResultArrowStream.cs
```diff
@@ -0,0 +1,421 @@
...
+
+    // Largest decimal precision guaranteed to fit each integer width: Int32 holds 9 full digits
+    // (max 2,147,483,647), Int64 holds 18 (max 9,223,372,036,854,775,807).
+    private const int MaxInt32Precision = 9;
+    private const int MaxInt64Precision = 18;
+
+    private const long NanosecondsPerSecond = 1_000_000_000L;
+
+    private readonly Ipc.IArrowArrayStream _inner;
+    private readonly List<ColumnTransform> _transforms;
```
CurtHagenlocher, line 77: I think there are probably some optimization opportunities in here. The C# builders aren't particularly efficient, and I think you could probably use the new `Int128` type in the BCL instead of `decimal` to load `Decimal128Array` more efficiently. We really need to get a bunch of this kind of functionality into the base C# Arrow libraries :/.

CurtHagenlocher (approved):
I've taken a pass over all the files and left a random assortment of comments. Looks good overall; I don't think there's anything that critically needs to be changed but I do have some suggestions for things that (I think :) would be improvements.

After the first two hours, I stopped focusing on the Snowflake-specific stuff and focused on the Arrow-specific code where I probably add more value anyway.

csharp/src/Native/Services/SnowflakeAccountUrl.cs
```diff
@@ -0,0 +1,54 @@
...
+            var port = network.Port != 443 ? $":{network.Port}" : string.Empty;
+            return $"{network.Protocol}://{network.Host}{port}";
+        }
+
+        var protocol = network?.Protocol ?? "https";
+        var host = account.Contains("snowflakecomputing.com", System.StringComparison.OrdinalIgnoreCase)
+            ? account
+            : $"{account}.snowflakecomputing.com";
+        var portSuffix = (network != null && network.Port != 443) ? $":{network.Port}" : string.Empty;
+        return $"{protocol}://{host}{portSuffix}";
```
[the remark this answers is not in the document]
ndglover: I've made some changes to make it more robust and take into account different region and cloud suffixes

csharp/test/Native/AdbcDrivers.Snowflake.Native.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageReference Include="xunit" Version="2.9.3" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="3.1.4">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Xunit.SkippableFact" Version="1.5.61" />
+    <PackageReference Include="FluentAssertions" Version="7.0.0" />
+    <PackageReference Include="NSubstitute" Version="5.3.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.0" />
+    <PackageReference Include="Microsoft.Extensions.TimeProvider.Testing" Version="9.0.0" />
```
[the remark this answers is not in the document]
ndglover: I'm leaving this for now, it's minor and will be better fixed when we pull the latest arrow packages with different transitive dependencies

csharp/src/Native/Services/ConnectionPool/ConnectionPoolManager.cs
```diff
@@ -0,0 +1,452 @@
...
+    /// <summary>
+    /// The secret that distinguishes one credential from another for the current auth type, or null
+    /// when the type carries no static secret (SSO / external browser).
+    /// </summary>
+    private static string? CredentialSecret(AuthenticationConfig auth) => auth.Type switch
+    {
+        AuthenticationType.UsernamePassword => auth.Password,
+        AuthenticationType.OAuth or AuthenticationType.Pat => auth.Token,
+        AuthenticationType.KeyPair => $"{auth.PrivateKey ?? auth.PrivateKeyPath} {auth.PrivateKeyPassphrase}",
+        _ => null,
```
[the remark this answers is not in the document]
ndglover: Created `CredentialFieldDelimiter `

csharp/src/Native/Configuration/AuthenticationConfig.cs
ndglover: Agreed, updated :-)

csharp/test/Interop/SnowflakeTestConfiguration.cs
ndglover: No, fixed

csharp/src/Native/Services/TypeConversion/SnowflakeDataType.cs
ndglover: Good spot, I included the DML types which never get returned. I took a closer look at the types and mapping and have made this more robust with added integration test,

csharp/src/Native/Services/TypeConversion/TypeConverter.cs
ndglover: Need to upgrade to v23, I'll tackle in subsequent PR

csharp/src/Native/Services/Authentication/SnowflakeLoginClient.cs
ndglover: Added the status code now

csharp/src/Native/Services/Query/ChunkedArrowArrayStream.cs
ndglover: I'd like to tackle this and some of the other performance related comments when we have updated to v23

csharp/src/Native/Services/Query/ChunkedArrowArrayStream.cs
ndglover: noted, will bucket with 'I'd like to tackle this and some of the other performance related comments when we have updated to v23'

csharp/src/Native/Services/Query/SnowflakeResultArrowStream.cs
ndglover: noted, will bucket with 'I'd like to tackle this and some of the other performance related comments when we have updated to v23'

lidavidm:
Thanks for this, I think we can address any followups in subsequent PRs, even if there's more feedback to add here.

PR #179, opened 2026-08-09
feat(go): add proxy-specific properties

davidhcoe:
## What's Changed

You *can* use a proxy today if you pass the `uri` property with the appropriate values like:

```
uri = "mysnowflakeuser:mysnowflakepass@myaccount?proxyHost=127.0.0.1&proxyPort=8080&proxyUser=myuser&proxyPassword=mypass&proxyProtocol=http"
```
This requires passing some dummy username/password, but is still valid for other auth types.

This PR adds the ability to provide the proxy values explicitly without having to rely on parsing the DSN value (which just so happens to be the first thing that is done today when `SetOptions` is called).

### DIFF
--- a/go/database.go
+++ b/go/database.go
@@ -530,13 +530,14 @@ func (d *databaseImpl) SetOptionInternal(k string, v string, cnOptions *map[stri
 	case OptionProxyHost:
 		d.cfg.ProxyHost = v
 	case OptionProxyPort:
-		d.cfg.ProxyPort, err = strconv.Atoi(v)
-		if err != nil {
+		port, parseErr := strconv.Atoi(v)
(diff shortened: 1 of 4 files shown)


lidavidm (approved):

CurtHagenlocher:
Given how common the need for proxying is, it would be nice to have a single set of property names standardized across as many connectors as possible. (This would preclude having the connector name as part of the property names.)

lidavidm:
> Given how common the need for proxying is, it would be nice to have a single set of property names standardized across as many connectors as possible. (This would preclude having the connector name as part of the property names.)

Do you want to file something upstream? I think this is good to have

CurtHagenlocher:
https://github.com/apache/arrow-adbc/issues/4672

